PR #57, opened 2026-08-13
Make agent_role fail-closed at both chokepoints (TODO L200)

seiya:
Closes the TODO.md high item: *"An unrecognized `agent_role` is INFERRED by one reader and SKIPPED by four others."* One field had **six** readers and four different fallbacks — one inferred the value, three skipped their own check, and the terminal filesystem-diff audit that `docs/AGENT_CONTRACT.md:22` calls authoritative returned early.

## What this changes

**Two chokepoints, because one is not enough.** The TODO's fix direction named record-launch alone. That cannot work: `_validate_actual_write_paths` and its caller read the **record-agent-run / finalize-child terminal payload**, a different input from the launch request. So:

- **Launch side** — `_validate_launch_request_payload` requires `agent_role` to equal `_required_child_agent_kind(step)`. The predicate was **not invented**: `validate_mcp_build_tool_invocation` and `_validate_run_gate_permissions` already demanded exactly this of the *capability*. It was extracted as `_require_child_agent_role_for_step` and applied one layer up, to the request that mints that capability — no new vocabulary, no new set literal.
- **Terminal side** — `AGENT_RUN_ROLES` (the complete `agent_runs.jsonl` vocabulary) and `WRITE_AUDITED_AGENT_ROLES` (the subset owning a write window) are named once; `record_agent_run` and its CLI-dispatch twin reject anything outside the former.
- **Normalization happens where it is read**, at the top of `prepare_launch_request_payload` — before that function renders the leaf prompt.
- `workflow_conductor.child_agent_role` now reads `STEP_REQUIRED_CHILD_AGENT` instead of restating it, and `_register_codex_thread` no longer defaults a missing role to `"substep"`.

## Severity, corrected a fourth time — in the safe direction

The entry flagged one unexecuted layer as "the next person's first step". It was executed. The Bash-detector gap is real (`_detect_bash_write_targets` returns no target for **7/7** of `cp`/`mv`/`dd of=`/`install`/`truncate`/`touch`/`ln`). **But a stray write *inside* `write_roots` is authorized for every role** by FS-diff containment, and every location bwrap rw-binds outside `write_roots` (`workspace/tmp/<arid>`, `gates/`, `hooks/`, `audit/`, `access_logs/`, the MCP log pins) is exempt for a legitimate role too.

**So under an enforced sandbox I found no write that an unrecognized role makes newly undetectable.** Not executed, and named as such: a host where `sandbox_enforced` is false. The fix still stands on its own — the defence was silently off, and "the sandbox covers it" is the second-layer argument this repo does not accept.

## Disclosure — the defects in this branch were mostly mine

Four review rounds; **rounds 1-3 each found the defect inside the previous round's fix**, with severity going *up* before it came down:

| Round | Defect | Whose |
|---|---|---|
| 1 | Fifth reader missed; both capability gates had no test at all | The original fix |
| 2 | Role validated but not canonicalized; the "proof" test was hollow | Round 1 |
| 3 | **The round-2 fix ran *after* the prompt was rendered** | Round 2 |
| 4 | Witness deleted; same defect on the sibling branch | Round 3 |

Round 3 is the one worth reading. My round-2 canonicalization landed 15 lines after `prepare_launch_request_payload` had already rendered the prompt, so it fixed nothing **and degraded the audit trail**: the persisted request said `substep` while the prompt beside it was rendered from `SUBSTEP` with no Task Card, meaning an auditor recomputing the card from the record obtains one that was never sent. Two reviewers found it independently.

Prose defects, all mine: the migration count was wrong twice (8, then 37 — a real measurement against a placement I then changed and never re-measured; it is **32 red over 28 payloads**); "nothing compared them" was **false** (`workflow_launch_check` did compare them, and unifying *removed* that live cross-check); `_write_allowed_output_manifest` was named as a harmed reader in three places and receives the launch role in none; and the test-class census was written three times and wrong three times.

**Two structural changes came out of that, rather than a fourth patch:**
- The census is no longer prose — it is `_DEFINITION_DRIVEN` / `_SAMPLE_DRIVEN` data asserted against the class's real methods, verified by renaming a test and watching it fail.
- The residual on the explicit-prompt branch is **closed**, not logged: a caller-supplied prompt paired with a non-canonical spelling is refused.

## Deliberately not in scope

The `_matches_phase_contract` set-equality gap (membership allowlist where the name says required set) is **split into its own TODO item**. A structural refactor riding on an enforcement change widens the review surface — the trade this repo has repeatedly declined.

Also kept, not deleted: the now-unreachable inference branches, and `required_child_agent_kind_mismatch`, which this branch made unreachable on the conductor path. "Unreachable" is a classification, and acting on one shipped fail-open here before.

## Verification

- Suite **4595 passed / 1 skipped**; `main` at `08dff50` measured **4580 / 1** → **+15**. (From a copied tree outside `$HOME` the absolute shifts by 5 checkout-depth `test_hooks_*` failures present on `main` too — compare the delta.)
- `ruff` per touched file identical to `main`: 3 / 1 / 45 / 6.
- Mutation checks across every round; the surviving hunks are docstring-only or behaviour-identical constant substitutions. Two **mechanism-level** deletions (the predicate never rejects; the vocabulary check removed from both terminal readers) are each caught, and a reviewer's independent 10-mutation sweep killed 10/10.

**Codex could not be used as a convergence signal**: two adversarial-review runs were launched and both stalled (46 min, then 28 in `verifying`) without returning a report. A stall is not a clean result, so the loop's upper stopping condition is not met — this stops on the class having dropped with a bounded remainder, which is recorded in TODO.md.


seiya:

---

- **An unrecognized `agent_role` is INFERRED by one reader and SKIPPED by four others, leaving the sandbox bound, the manifest empty, and the terminal write audit silently disabled** (2026-08-13, **high** — reproduced on `origin/main` and unchanged by the branch that found it). Five readers of one field, three different fallbacks:
  1. `build_capability_document` (`tools/orchestration_runtime.py:4721`) INFERS: lines 4730-4736 turn `agent_role="bogus"` into `substep` when `substep` is present, minting a normal capability whose `write_roots` are the IR run dir WITH its trailing slash.
  2. `_allowed_output_paths_for_launch` (`:6531`) SKIPS the phase-contract check for an unrecognized role and returns those write_roots through `_normalize_rel_posix`, which STRIPS the slash. The write hook grants directory authority only to entries ending in `/` (`tools/hooks/common.py:4188-4201`), so the manifest's one entry matches nothing: measured on the captured `compile_generate.request.json`, `Write` to `spec.ir.yaml`, `ir_meta.json` and an arbitrary `evil.md` are ALL blocked.
  3. `_validate_child_write_contract_preflight` (`:7545`) SKIPS.
  4. `build_bwrap_profile` (`:7979-8000`) reads the CAPABILITY document, where the slash survives — so the sandbox rw-binds the whole IR run dir exactly as for a legitimate substep. The filesystem is not locked down; only the file-tool manifest is.
  5. `_validate_actual_write_paths` (`:9585-9591`) and its terminalization caller (`:14585-14591`) both RETURN EARLY for a role outside `{orchestration, step, substep}`, so the FS-diff attribution `docs/AGENT_CONTRACT.md:22` names as authoritative never runs. Measured with a terminal status: `role="substep"` reaches the check (`ValueError: run write baseline missing`), `role="bogus"` returns without validating.
  - **Net effect, stated as measured**: the leaf cannot author its own deliverables (2), the sandbox still grants it the directory (4), and the audit that would flag an unmanifested write there is off (5). Rated high because a backstop this repository designates authoritative is silently disabled — that is a removed defence, not a hypothesis. **NOT executed, and therefore not claimed**: whether a write can actually land in that bound directory, which needs a write path the file-tool hook does not intercept (the Bash-detector gaps in the medium-severity audit item below are the candidate). Anyone picking this up should run that step before rating the exploitability.
  - **How this entry's severity was wrong three times, which is the transferable part.** Draft 1: medium, from an unexecuted premise citing a function that does not exist. Draft 2: high, from a reproduction that stopped at the manifest PRODUCER. Draft 3: medium, from running the manifest CONSUMER (the hook) and concluding lockout — while never opening the sandbox binder or the FS-diff. Each draft executed one more layer and asserted the rest. The rule that would have prevented all three: **enumerate every reader of the field first, then execute each one**; a severity claim is finished only when no reader is left unopened.
  - **The second, milder gap in the same function**: `_matches_phase_contract` ends every phase branch in `path in <set>`, so it is a membership allowlist, not the required set `compile_required`'s name implies — a `compile.generate` request declaring ONLY `ir_meta.json` is accepted although the conductor (`workflow_conductor.build_launch_request`) always declares both. Under-declaring costs the declarer authority rather than gaining any. `agent_role="step"` is likewise accepted on a compile payload although `child_agent_role()` assigns `substep` to every compile launch and `step` only to `build`.
  - Fix direction: normalize `agent_role` ONCE, where the payload is validated, using the same inference `build_capability_document` performs, so all three readers see one value; then key the expected set on `(agent_role, step, substep)` and require set EQUALITY per substep (`compile.generate` = both files, `compile.verify` = `ir_meta.json`, `compile.static` = `compile_static_meta.json`). Do NOT leave the skip-on-unknown-role branch reachable: an unrecognized role must fail closed, not fall through to `write_roots` — but budget for the migration, since several `record_launch` tests omit `agent_role` entirely and currently depend on the inference (making the unknown role fail closed turns 8 of them red). Deriving each substep's set from ONE named constant is what finally makes it testable: until the set has a single definition, a test outside the function can only SAMPLE names it must reject, so adding one more name to `compile_required` is invisible to the suite (demonstrated escapes, all green against both pins: one more literal name `<ir_ref>/summary.md` in the set; a subtree prefix `<ir_ref>/scratch/`; an extension family `path.endswith('.log')` at the IR root. The probes those tests carry are shape-samples — a name, a subtree, a directory entry — and three review rounds each broke the previous sample's shape, which is the evidence that sampling is the wrong instrument here.)
  - Completion criterion: a payload whose `agent_role` the conductor never assigns to that phase is rejected at record-launch (not silently normalized into one path and skipped in another), and a request that omits a declared deliverable is rejected — with a test per compile substep, driven by the captured production payloads under `tools/tests/data/conductor_launch_requests/`, and one that drives `build_capability_document` -> `_allowed_output_paths_for_launch` in sequence the way `record_launch` does, since testing either alone is what hid this.
  - **DONE (both chokepoints closed, 2026-08-13).** Commits `953bc05` (launch) + `f9c79a7` (terminal). The fix direction above was written for record-launch alone; **that is not sufficient, and the reason is the reusable part**: readers 5 and its caller read the record-agent-run / finalize-child TERMINAL payload, a different input from the launch request, so normalizing the launch payload cannot reach them. Two chokepoints were needed. Enumerating the readers first (the rule this entry itself proposed) also turned up a **sixth**: `record_launch:17451-17458` has its own fallback chain — a present-but-bogus role is taken verbatim, an absent one is derived from `step`, and an unknown `step` yields the literal `"unknown"`, written to `session_run_index`.
    - **A claim in this branch's commit messages is WRONG, and a blank-slate reviewer caught it.** `953bc05` says of the conductor's `child_agent_role` and the runtime's `STEP_REQUIRED_CHILD_AGENT`: "Two definitions of one fact were survivable while nothing compared them." **Something did compare them.** `workflow_launch_check` took the conductor's answer as `require_child_agent` and reported `required_child_agent_kind_mismatch` when it disagreed with the table; the conductor fed it `child_agent_role(phase)`. Unifying the two REMOVED that live comparison rather than adding one. The unification still stands — drift made IMPOSSIBLE beats drift REPORTED — but two consequences have to be stated rather than glossed: (1) the launch-side role check is a **tautology for every conductor-produced request** (`x != x`), so its whole value is against a non-conductor caller of the `record-launch` CLI, which is exactly the threat this entry described; and (2) `required_child_agent_kind_mismatch` now has no reachable producer on the conductor path. It is NOT deleted — "unreachable" is a classification, and this repository has shipped fail-open by acting on one — but it is dead weight the next person should decide about deliberately.
    - **What was built.** The required predicate ALREADY EXISTED: `validate_mcp_build_tool_invocation` and `_validate_run_gate_permissions` both demand `capability.agent_role == _required_child_agent_kind(step)`. It was extracted as `_require_child_agent_role_for_step` and applied to the launch request that MINTS the capability, so there is no new vocabulary and no new set literal. On the terminal side, `AGENT_RUN_ROLES` (the complete `agent_runs.jsonl` vocabulary) and `WRITE_AUDITED_AGENT_ROLES` (the subset owning a write window) are named once and read by the four sites whose question they actually answer (the two terminal-audit early returns, `record_agent_run`, and its CLI-dispatch twin). Three verbatim copies of the same three members REMAIN on purpose — two in `build_capability_document` ("which roles may a capability name") and one in the `repair_legacy_agent_runs` backfill ("which rows represent an agent that ran"); they coincide with the write-audited set today but are not defined by it, and a comment at the constant says so. An earlier draft of this annotation claimed every copy had been folded in. `workflow_conductor.child_agent_role` now reads `STEP_REQUIRED_CHILD_AGENT` instead of restating `"step" if step == "build" else "substep"` — two definitions of one fact stop being survivable once record-launch rejects the disagreement.
    - **The unexecuted layer this entry flagged, now executed — and the severity was wrong a FOURTH time, in the safe direction.** The entry said a write landing in the rw-bound directory needed "a write path the file-tool hook does not intercept". Confirmed that such a path exists: `_detect_bash_write_targets` (`tools/hooks/cli.py`) returns NO target for `cp` / `mv` / `dd of=` / `install` / `truncate` / `touch` / `ln` (7/7 executed), and `cli.py:1632` returns `common_decision` when there are no targets, so the write guard never runs. **(That measurement was of `origin/main` at the time; item 378(d) has since closed it — all seven now produce a target. The reasoning below about what the audit would have caught is unaffected, because it never rested on the extractor.)** **But the audit would not have caught that write anyway**: a stray file INSIDE `write_roots` is authorized for every role by FS-diff containment (`docs/AGENT_CONTRACT.md:22`) — executed for `substep` / `step` / `bogus`, all three clean. The role only changes the verdict for writes OUTSIDE `write_roots` (`substep`/`step` FLAGGED, `bogus` clean), and enumerating every location `build_bwrap_profile` rw-binds outside `write_roots` — `workspace/tmp/<arid>`, `gates/<arid>/`, `hooks/`, `audit/`, `access_logs/<arid>.jsonl`, the `mcp_owned_audit_logs` file pins — every one is exempt or authorized for a legitimate role too. **So under an enforced bwrap sandbox no write was found that the bogus role makes newly undetectable.** NOT executed: a host where `sandbox_enforced` is false, and whether a leaf can reach `record-launch` at all. The defence was still silently off, and "the sandbox covers it" is the second-layer argument this repository does not accept, so the fix stands on its own.
    - **Two probes decided what NOT to build.** (a) Swapping a step child's terminal role to `orchestration` to get a different FS-diff baseline does not weaken the audit — it flags MORE, because the orchestration branch requires every changed path to be declared in `output_refs`; measured all three ways. (b) `skipped_by_checkpoint` is in the vocabulary but not write-audited, which would be an evasion if it could also be terminal — it cannot, since it requires `status=skipped` and `skipped` is not in `TERMINAL_STATUSES`. No launch-role cross-check was added, and the reason is an execution record rather than a judgement.
    - **Completion criterion, both halves.** A payload whose `agent_role` the conductor never assigns to that phase is rejected AT record-launch, before the session-run-index row is appended (so it adds no orphan `running` row to the audit's item (f)); and the terminal payload is rejected rather than silently skipping the write audit. Seven pins, all generated by ITERATING `STEP_REQUIRED_CHILD_AGENT` / `AGENT_RUN_ROLES` rather than sampling names — the property the phase-contract tests could not express from outside the set — plus one driving the seven captured production payloads through the real `_validate_launch_request_payload`, and one driving `build_capability_document` -> `_allowed_output_paths_for_launch` in record-launch's own order (testing either alone is what hid this).
    - **The migration figure in this entry was wrong, and so was my correction of it — measured a third time, it is 32 red tests over 28 payloads.** The entry recorded "8 tests turn red", which counted payloads reaching `build_capability_document` with no role — a different predicate. I then published **37**, which was the failure count from an EARLIER placement of the check (before the deterministic-flag and `agent_model` checks); I moved the check to preserve those tests' error ordering and never re-measured. Re-measured on the final placement, by dropping only the two changed source files onto `main`'s tree: **32 tests turn red — 22 on a missing `agent_role`, 10 on a role the step does not take.** The fixture side is **28 launch payloads**: 28 gained `agent_role`, 4 had it corrected (`step` -> `substep` on `generate` and on `validate`), 20 gained `allowed_output_paths` — that third class **never appears as a per-test count**, because a missing output only becomes an error once the role is supplied, which is itself the shape of the gap: with an unknown role `_allowed_output_paths_for_launch` returned before its own "requires explicit allowed_output_paths" check. The transferable part is not the number: **a figure measured against a draft you then changed is a stale figure, and re-measuring is not optional just because the first measurement was real.**
    - **Mutation check, across three passes as the branch grew.** The terminal commit alone: 5 hunks, 3 killed; one survivor was real (the library-side rejection in `record_agent_run` was unpinned while only its CLI-dispatch twin was covered — "helper pinned, handler not") and got a test. Branch-wide: 10 hunks, 6 killed; the 4 survivors were all behaviour-preserving substitutions, and the two capability-side ones exposed that `validate_mcp_build_tool_invocation` / `_validate_run_gate_permissions` had **no test at all** for a mismatched capability role — pinned in round 1. Round-1 hunks: 2 of 2 killed. Two MECHANISM-level deletions (the predicate never rejects; the vocabulary check removed from BOTH terminal readers) are each caught, by 4 and 2 tests — hunk reverting cannot see a mechanism that is live but unobserved, which is why they were run separately. An independent reviewer's 10-mutation sweep killed 9; the survivor was `child_agent_role` restating the literal, now pinned on the only behaviour that distinguishes it (an unsupported step raises instead of silently answering `substep`).
    - Suite **4595 passed / 1 skipped** at the branch tip; `main` at `08dff50` measured **4580 / 1** on this checkout, so **+15**. The figure moved with every review round (4587 -> 4589 -> 4591 -> 4592 -> 4595) and was stale in this file twice; treat any number written here as of its round. `ruff` per touched file identical to `main` (3 / 1 / 45 / 6). Re-running from a copied tree outside `$HOME` shifts the absolute by 5 checkout-depth `test_hooks_*` failures present on `main` too — compare the DELTA.
    - **Round 4 is where the class finally dropped: zero functional defects from the blank-slate reviewer (10/10 mutations killed, 3 of them mechanism-level), and the targeted reviewer's two findings were both bounded.** (a) Round 3 kept the validator write-back as an explicit "backstop for a caller that skips prepare" and DELETED the only test that drove the validator without prepare — neutering the assignment left the whole suite green. Keeping a mechanism while deleting its witness is this branch's own recurring class one layer up; the witness is restored. (b) The round-3 defect survived verbatim on the OTHER branch of `prepare_launch_request_payload`: when the caller supplies its own prompt nothing is re-rendered, so a prompt built from `"SUBSTEP"` shipped while the persisted request recorded `"substep"` — the same record-vs-prompt divergence, one branch over. Closed by REFUSING that pair (caller-supplied prompt + non-canonical spelling), which is narrow by construction and unreachable from the conductor, rather than recorded as a residual.
    - **The test-class census is no longer prose.** It was written three times and wrong all three (all-seven-iterate when five did; a test name a rename had already deleted; "eight tests" when there were nine). Restating a census is the wrong shape — it goes stale silently on any add or rename. It now lives in `_DEFINITION_DRIVEN` / `_SAMPLE_DRIVEN` data asserted against the class's actual methods by `test_this_class_census_is_accurate`, verified by renaming a test and watching it fail. The distinction it tracks is the one this whole entry keeps rediscovering: a test that ITERATES a definition extends automatically, a test that SAMPLES shapes can show a rejection but never set identity.
    - **Codex could not be used as a convergence signal on this branch.** Two adversarial-review runs were launched; both stalled (46 min, then 28 min in `verifying`) and neither returned a report. The second produced only an early partial verdict — `needs-attention`, naming the same validator-mutation drift path the subagents found — before stalling. A stall is not a clean result, so the skill's upper stopping condition (security-clean twice AND Codex clean) is NOT met here, as it has not been in any recorded loop.
    - **Round 3 found the round-2 fix was placed AFTER the state it guarded, and it is the worst self-inflicted defect of the branch.** Round 2 canonicalized `agent_role` inside `_validate_launch_request_payload` to stop `"SUBSTEP"` producing an empty Task Card. But `record_launch` calls `prepare_launch_request_payload` FIRST, and that is what renders `launch_prompt_full` (embedding `_build_task_card`, which reads the role with `.strip()` and no `.lower()`); the canonicalization ran 15 lines later. So the symptom was not fixed — and the record got WORSE: the persisted `launches/<arid>.request.json` said `substep` while the prompt beside it had been rendered from `SUBSTEP` with no card, meaning an auditor recomputing the card from the request obtains one that was never sent. Two independent reviewers found it; the test that "proved" the round-2 fix was hollow because it called `_build_task_card` itself after mutating the payload instead of reading the prompt `record_launch` would persist. Fixed by normalizing at the TOP of `prepare_launch_request_payload`, before the render (normalization only — an unknown role is left untouched for the validator to reject, so it can never launder a bad value); the test now drives prepare -> validate -> extract and asserts on the RENDERED TEXT, and was verified to fail against the round-2 code.
    - **Three more of my own claims were false and are corrected in place.** (a) `_write_allowed_output_manifest` was named in a comment, a test docstring and a commit message as a reader harmed by an unlowered role; it is not — record-launch calls it with no `agent_role`, and its other call sites pass the literal `"orchestration"`. (b) `_normalized_agent_role`'s docstring said "every reader of the field normalizes this way"; four do not, which is precisely why canonicalizing at the chokepoints rather than trusting readers is the right shape. (c) A comment said "six downstream readers" and then listed five — the sixth is `_register_codex_thread`, closed in this same branch. Also pinned this round: the write-audit USE SITE reads the narrower constant (a reviewer's mutation swapping it for `AGENT_RUN_ROLES` had survived the whole suite — pinning the definitional difference between the two sets was not the same as pinning which one the early return consults).
    - **NOT closed by this work**: the set-equality gap in `_matches_phase_contract` (the "second, milder gap" bullet above), deliberately left out so a structural change does not ride on an enforcement change. It is now its own item below.

PR #58, opened 2026-08-13
Take the calibration test's input out of the operator's workspace

seiya:
Closes the second half of TODO L242. The first half — a retention policy for `workspace*` — is **rejected by decision, not deferred**: those are the gate's execution workspace and how they are used is the operator's call, not something this repository legislates. No `docs/RUNBOOK.md` change, and **zero bytes deleted** (`du -sb workspace workspace_*` = `15,040,129,410` before and after).

## The defect

`test_real_full_fidelity_predicate_set_is_not_degenerate` pinned its input at `<repo>/workspace/ir/.../spec.ir.yaml` and called `skipTest("real IR artifact not present in this checkout")` when it was missing. `workspace/` is gitignored, so on a fresh clone that test could never run. It skipped silently for weeks — the suite's standing "1 skipped" — while its own comment declared it the real-shape calibration line.

The item located the cause in the missing retention rule. It is not: the cause is a test reading the operator's execution workspace.

## What this does

- Captures the real IR **byte-for-byte** (`cmp`-verified, sha256 `b6a1352d…f15e`, 23,835 bytes) into `tools/tests/data/real_ir/` and drops the `skipTest`. A missing fixture is repository corruption, so it fails.
- Adds `tools/tests/test_skip_reasons_are_declared.py`: every declared skip in the tracked corpus must carry a string-literal reason listed in one table of twelve host capabilities. The retired reason string is the probe driving the reject path.
- The calibration test drives the same fixture twice — as captured, and with its pass predicates rewritten to `verdict.overall` — because a clean run alone is also what an unreadable document produces.

**Measured:** `origin/main` collects 4596 and runs 4595 passed / 1 skipped; the tip runs **4610 passed / 0 skipped**. `ruff check tools/ mcp_servers/` = 83 errors, unchanged.

## Disclosure

Seven review rounds (two subagents per round, Codex three times). Almost every finding was inside the previous round's fix.

- **The first guard answered the wrong question.** It AST-scanned path expressions and required repo-root-anchored paths to be tracked. Three reviewers produced ~20 evasions, two of them ordinary style already in the tree; adding a module that read `workspace/` and skipped, with `.resolve()` on its own line, left it green. Replaced wholesale rather than patched. Earlier in its life it also produced **382 false positives**, and its own sweep was unpinned — replacing the module glob with a single file left every test passing.
- **The replacement did not converge either, and this PR stops by decision.** Rounds 5-7 each produced 3-6 findings, all the same shape: one more Python binding form the static resolution did not model. No descent in severity across the last three. **None of those constructs exists in the corpus** — all 23 skip sites use `unittest.skipUnless` or `self.skipTest` — so every finding after round 4 was written by a reviewer hunting for it. The module docstring and TODO now say plainly that this is regression prevention against ordinary spellings, not enforcement. The instrument for the stronger claim is a runtime skip inventory, recorded as its own TODO item.
- **Stopping conditions not met.** Codex returned `needs-attention` on all three launches and never went clean. The two proxies disagreed: independent mutation sweeps (170 and 193 mutants) recommended stopping in rounds 4 and 6, while the blank-slate reviewer found live routes both times — a sweep can only mutate mechanisms that exist, and every round after 3 found a missing one.
- **Defects I introduced and had to fix:** the 382 false positives; an assertion that comprehended over an AST with zero matching nodes and so could not fail; a calibration test that passed on `{}` and on unparseable YAML; **a false sentence in a commit message** (I read a test failure without reading which path it named); **suite figures wrong three times**, the third "correcting" a right number with a wrong one; and **a broken history** — a `git commit --amend` meant to fix a message silently absorbed a staged file replacement, so a commit claimed work that `git log -S` could not find. That was repaired by squashing, with the reason recorded in the message.


seiya:

*(Misfiled. Review round 3 of the issue #181 pull request established that the two `Generate.static` call-site blocks were collapsed in PR #60 (`6247fa3`, merged `4bbeece`); the item's "PR #58" describes a DIFFERENT item's pull request. The record for this item is https://github.com/seiya/atmofab/pull/60#issuecomment-5560285997. The verbatim body below is unchanged.)*

---

- **Collapse the two near-identical `Generate.static` gate call-site blocks** (2026-08-12, medium — structural). `_validate_generate_outputs` and `_validate_generate_outputs_for_generation` (`tools/validate_pipeline_semantics.py`) carry near-identical bodies: the same `index(case_id)` scan, the same literal-metric floor, and the same three `problem` model gate invocations, differing only in how they reach `src_dir`. Every rule change in this area has to be made twice, and "fixed one copy, missed the twin" is a repeat failure mode in this repository — the 2026-08-12 continuation work had to assert its one new rule at both entry points to be sure.
  - Fix direction: extract the shared body to one helper taking `(repo_root, execution, src_dir, violations)`; leave the two callers as the two ways of resolving `src_dir`.
  - Completion criterion: one definition of the scan-and-gate sequence remains; the existing tests of both entry points stay green without being duplicated further.
  - Deliberately NOT done with the gate fixes that exposed it: a structural refactor riding on an enforcement change widens the review surface, which is the trade this repository has repeatedly decided against.
  - **DONE (2026-08-13).** Suite **4616 passed** on both `origin/main` (`deef342`) and the branch — this change adds and removes no tests, so an unmoved count is the point. Note the standing "1 skipped" that earlier DONE notes in this file report is **gone**: the L242 work (PR #58) took the calibration test's input out of the operator's workspace, so the baseline is a clean 4616 with no skip. `ruff check tools/ mcp_servers/` = 83 errors on both revisions, identical histogram (35 E702, 19 F401, 12 F841, 5 F541, 5 E741, 4 F821, 3 E401); the two touched files carry 1 E741 between them, unchanged.
    - **The fix direction above was written on a wrong premise, and the correction is the reusable part.** It says "extract the shared body to one helper … leave the two callers as the two ways of resolving `src_dir`", which assumes each of the two functions carries its own resolution prologue. Only `_for_generation` does; the other is handed an already-resolved `src_dir` by its caller (`_execution_in_scope_src_dir`, via `_validate_impl`). And `_validate_generate_outputs`'s body **was** the shared block exactly — 4691-4762, ending with nothing of its own — so it already **was** the helper the direction asked for, at that exact signature. Extracting a second one would have left it a one-line passthrough. Landed instead as: `_validate_generate_outputs` keeps its signature and its body untouched and widens its return to `tuple[list[Path], list[str]] | None`; `_for_generation` keeps its `gen_dir`/`src_dir` prologue and calls it. The 72 duplicated lines appear in the diff only as deletions.
    - **Why the return had to widen rather than stay `-> None`.** `_for_generation`'s tail (runner sources, checks module, dep-op-on-model-files, infra signatures, component surface, OpenMP floor) consumes `model_files` and `dep_spec_ids`, both computed inside the shared block, and the block's one early return (absent/mis-named model source) must skip that tail. So the shared part has to hand both values back and signal that early return — hence `None`, which reproduces the old `if not model_files: return` exactly, with the violation still appended inside the callee.
    - No new tests. `WrappedLiteralMetricAssignmentTest` already drives both entry points behaviorally and becomes the wiring pin for the new structure unmodified; only its docstring changed, which had stated the duplication as an explicit contract and would otherwise now be false. `test_reaches_the_post_generate_sibling_list` still pins the tail, which this change keeps verbatim.
    - **Observed and deliberately left**: `_validate_post_build_stage_impl` (`tools/validate_pipeline_semantics.py`) carries a third copy of the Makefile trio (`_validate_fortran_makefile_src_dir` + `_validate_makefile_test_no_relink` + `_validate_makefile_test_invokes_cases`) on its own `src_dir`. That is the same three checks in a different stage rather than a copy of the scan-and-gate sequence this item names, and folding it in would change what `post_build` validates — out of scope here.

PR #60, opened 2026-08-13
Collapse the duplicated Generate.static scan into one definition (TODO L195)

seiya:
Closes TODO.md L195 (medium, structural).

## Why

`_validate_generate_outputs` and `_validate_generate_outputs_for_generation` (`tools/validate_pipeline_semantics.py`) carried **72 byte-identical lines**: the model-file lookup and its not-found early return, `_component_dep_spec_ids`, the per-model `_joined_masked_fortran_view`, the `index(case_id)` scan, the literal-metric floor, the three `problem` model gate invocations, and the Makefile checks.

Every rule change in this area had to be made twice, and "fixed one copy, missed the twin" is a recorded repeat failure here — the 2026-08-12 continuation work had to assert its one new rule at both entry points to be sure, and the test class covering that stated the duplication as an explicit contract in its own docstring.

This is sequenced deliberately ahead of **L174** (`function` / `module procedure` reachability, the last remaining high), which must add a rule to exactly this pair. Doing it now means that rule gets written once.

## The TODO's fix direction rested on a wrong premise

It said: extract a helper taking `(repo_root, execution, src_dir, violations)`, leaving the two callers as the two ways of resolving `src_dir`. Two facts do not fit:

1. **Only `_for_generation` resolves its own `src_dir`.** The other is handed one by `_validate_impl` via `_execution_in_scope_src_dir`.
2. **`_validate_generate_outputs`'s body WAS the shared block exactly** — 4691-4762, ending with nothing of its own — at the exact signature the direction asked for. So it already *was* that helper; extracting a second one would have left it a one-line passthrough.

Landed instead as: `_validate_generate_outputs` keeps its signature and its body untouched and widens its return; `_for_generation` keeps its `gen_dir`/`src_dir` prologue and its tail verbatim, and calls it. The 72 lines appear in the diff **only as deletions**.

## Why the return had to widen rather than stay `-> None`

`_for_generation`'s tail — runner sources, checks module, dep-op-on-model-files, infra signatures, component surface, OpenMP floor — consumes `model_files` and `dep_spec_ids`, both computed *inside* the shared block, and the block's one early return (absent/mis-named model source) must skip that tail. So the shared part has to hand both values back and signal that early return. `-> tuple[list[Path], list[str]] | None`, with `if scanned is None: return` reproducing the old `if not model_files: return` exactly — the violation is still appended inside the callee.

## No rule change, no new test

`WrappedLiteralMetricAssignmentTest` already drove both entry points behaviorally and becomes the wiring pin for the new structure **unmodified**. Only its docstring changed: it had stated the duplication as a contract and would otherwise now be false. `test_reaches_the_post_generate_sibling_list` still pins the tail, which this change keeps verbatim.

## Disclosures

- **A third copy of the Makefile trio exists and is deliberately left.** `_validate_post_build_stage_impl` invokes `_validate_fortran_makefile_src_dir` + `_validate_makefile_test_no_relink` + `_validate_makefile_test_invokes_cases` on a `src_dir` it resolves itself. That is the same three checks in a different stage rather than a copy of the scan-and-gate sequence this item names, and folding it in would change what `post_build` validates. Recorded in TODO.md, and — after review round 1 — **stated in the docstring too**, which is where a maintainer changing a Makefile rule would actually read it. The first draft claimed "single definition" over a set including the Makefile checks; that claim was too broad.
- **The standing "1 skipped" figure in this file's earlier DONE notes is gone.** PR #58 took the calibration test's input out of the operator's workspace, so the baseline is a clean 4616 with no skip. Corrected in TODO.md so the next reader does not carry the stale figure forward.

## Verification

| Check | `origin/main` (`deef342`) | Branch tip |
|---|---|---|
| `python3 -m pytest tools/tests -q` | 4616 passed, 0 skipped | **4616 passed, 0 skipped** |
| `ruff check tools/ mcp_servers/` | 83 errors | **83 errors**, identical histogram |
| `index(case_id)` scan sites | 2 | **1** |
| literal-metric floor sites | 2 | **1** |

No test was added or removed, so the **unmoved** count is the evidence, not a rising one. Histogram both sides: 35 E702, 19 F401, 12 F841, 5 F541, 5 E741, 4 F821, 3 E401; the two touched files carry 1 E741 between them, unchanged.

One review round (medium) ran against the diff and found no correctness bug. It verified the collapse independently rather than reading the claim: an AST source-segment diff confirming the removed lines are byte-identical to the retained body, a name-scan confirming the tail loads only `model_files` / `dep_spec_ids` with nothing leaking out of the deleted loop, the return annotation checked against `_model_files_in_src_dir` / `_component_dep_spec_ids`, every caller checked against the widened return, and its own full-suite run. Its one finding — the over-broad docstring claim — is fixed in `feb0fba`.

**Not run: Codex.** Only the subagent lens has seen this diff.


seiya:
**Migrated from `TODO.md` (issue #181).** This item was recorded in `TODO.md` until commit `f88ad94` (main, 2026-09-06) and is reproduced below verbatim; the pull request for issue #181 removes it from that file. It was first filed on PR #58 by a destination rule that takes the first `PR #N` in the item body; review round 3 established that the two `Generate.static` call-site blocks were collapsed in PR #60 (`6247fa3`, merged `4bbeece`); the item's "PR #58" describes a DIFFERENT item's pull request, so the record belongs here. Line-number citations (`L<n>`) and "above" / "below" references are as they stood in that file.

---

- **Collapse the two near-identical `Generate.static` gate call-site blocks** (2026-08-12, medium — structural). `_validate_generate_outputs` and `_validate_generate_outputs_for_generation` (`tools/validate_pipeline_semantics.py`) carry near-identical bodies: the same `index(case_id)` scan, the same literal-metric floor, and the same three `problem` model gate invocations, differing only in how they reach `src_dir`. Every rule change in this area has to be made twice, and "fixed one copy, missed the twin" is a repeat failure mode in this repository — the 2026-08-12 continuation work had to assert its one new rule at both entry points to be sure.
  - Fix direction: extract the shared body to one helper taking `(repo_root, execution, src_dir, violations)`; leave the two callers as the two ways of resolving `src_dir`.
  - Completion criterion: one definition of the scan-and-gate sequence remains; the existing tests of both entry points stay green without being duplicated further.
  - Deliberately NOT done with the gate fixes that exposed it: a structural refactor riding on an enforcement change widens the review surface, which is the trade this repository has repeatedly decided against.
  - **DONE (2026-08-13).** Suite **4616 passed** on both `origin/main` (`deef342`) and the branch — this change adds and removes no tests, so an unmoved count is the point. Note the standing "1 skipped" that earlier DONE notes in this file report is **gone**: the L242 work (PR #58) took the calibration test's input out of the operator's workspace, so the baseline is a clean 4616 with no skip. `ruff check tools/ mcp_servers/` = 83 errors on both revisions, identical histogram (35 E702, 19 F401, 12 F841, 5 F541, 5 E741, 4 F821, 3 E401); the two touched files carry 1 E741 between them, unchanged.
    - **The fix direction above was written on a wrong premise, and the correction is the reusable part.** It says "extract the shared body to one helper … leave the two callers as the two ways of resolving `src_dir`", which assumes each of the two functions carries its own resolution prologue. Only `_for_generation` does; the other is handed an already-resolved `src_dir` by its caller (`_execution_in_scope_src_dir`, via `_validate_impl`). And `_validate_generate_outputs`'s body **was** the shared block exactly — 4691-4762, ending with nothing of its own — so it already **was** the helper the direction asked for, at that exact signature. Extracting a second one would have left it a one-line passthrough. Landed instead as: `_validate_generate_outputs` keeps its signature and its body untouched and widens its return to `tuple[list[Path], list[str]] | None`; `_for_generation` keeps its `gen_dir`/`src_dir` prologue and calls it. The 72 duplicated lines appear in the diff only as deletions.
    - **Why the return had to widen rather than stay `-> None`.** `_for_generation`'s tail (runner sources, checks module, dep-op-on-model-files, infra signatures, component surface, OpenMP floor) consumes `model_files` and `dep_spec_ids`, both computed inside the shared block, and the block's one early return (absent/mis-named model source) must skip that tail. So the shared part has to hand both values back and signal that early return — hence `None`, which reproduces the old `if not model_files: return` exactly, with the violation still appended inside the callee.
    - No new tests. `WrappedLiteralMetricAssignmentTest` already drives both entry points behaviorally and becomes the wiring pin for the new structure unmodified; only its docstring changed, which had stated the duplication as an explicit contract and would otherwise now be false. `test_reaches_the_post_generate_sibling_list` still pins the tail, which this change keeps verbatim.
    - **Observed and deliberately left**: `_validate_post_build_stage_impl` (`tools/validate_pipeline_semantics.py`) carries a third copy of the Makefile trio (`_validate_fortran_makefile_src_dir` + `_validate_makefile_test_no_relink` + `_validate_makefile_test_invokes_cases`) on its own `src_dir`. That is the same three checks in a different stage rather than a copy of the scan-and-gate sequence this item names, and folding it in would change what `post_build` validates — out of scope here.

PR #61, opened 2026-08-14
Take the three problem-model gates' structure from a parser, not a regex walk

seiya:
Stage A of two. Replaces the hand-rolled regex walk that located Fortran procedure bodies for the
three `problem` model gates with a parser front end (`tools/fortran_structure.py`, tree-sitter-fortran).
**Envelope semantics are unchanged** — that is the merge condition, and it is measured, not asserted.

> **Stacked PR — do not merge alone.** Stage B (`feat/fortran-function-visibility`) sits on top of
> this branch and contains, besides the visibility change, the fixes for defects found by review
> **in this branch's code**: a `RecursionError` on deeply nested source that surfaced as
> `schema_load_failed` and discarded every other violation, a classification channel a leaf could
> forge through its own filename, refusals of legal sources carrying statement labels, and a silent
> failure mode if the grammar renames a node type. Merging this alone ships those. See the notes at
> the bottom.

## What changed

`_fortran_subroutine_envelopes` is now an adapter over `fortran_structure.parse_view`. The walk it
replaces was rewritten four times and broken sixteen, always the same way: a spelling the language
allows and the rules did not enumerate — the one-word `endsubroutine`, a bare `end`, a construct
named after a keyword, a variable named after a keyword, an `interface` body's own `end subroutine`.
Each was fail-open and silenced all three gates for a whole file. That set is not closed by
enumeration, so the enumeration is gone rather than extended.

The three gates no longer each parse the same file: `_validate_generate_outputs` computes the
envelopes once and passes them, and the `intent(out)` set — three copies of one pattern before — is
computed once, in the envelope.

## Evidence

| measurement | result |
|---|---|
| 365 in-tree `*_model.f90`, all 894 envelope tuples + all 69 violations vs `origin/main` | **byte-identical** (literal-outputs 38, dependency-dataflow 27, metric-only-kernel 4, 0 refused) |
| flang cross-check (`-fc1 -fdebug-unparse-no-sema`, Debian flang 21.1.8) | **365/365 agree** |
| suite | 4630 passed / 0 skipped (`origin/main` 4616) |
| `ruff check tools/ mcp_servers/` | 83 on both revisions, identical histogram |
| mutation check | 14/14 mechanism-level killed; 16 hunks → 11 killed, 5 survivors (3 justified, 2 were real gaps and are fixed in this branch) |

`tools/fortran_structure_differential.py` is that listing: a development harness, run at two
revisions and diffed, never a suite test. Enumerating the corpus needs `os.walk`, not `grep -r` —
`grep` is a ripgrep alias here and sees 1 of the 365 files.

## What this costs, measured

A parse carrying an ERROR/MISSING node is refused (a content violation naming the rewrite) rather
than read partially. Of 48 rows assigning to a variable named after a keyword, **38 are still
analysed and flagged and 10 are refused**; of 35 construct-name rows, 20 analysed and 15 refused.
Every refused row is a legal program that the parser lexes as the statement its name spells, and
every one is repairable by a rename, which the message asks for. Both sets are pinned as SETS, so a
change that refuses one more row fails a test rather than surfacing as a refused Generate in a
billed run.

Inputs `gfortran -std=f2008` REJECTS (a stray `end do`, an unmatched `end function`, an unterminated
body or interface) are refused too, where the walk read them deliberately long. None is reachable
for a fortran node: `_gate_inproc` runs the static check only when lint AND syntax passed.

## Attribution

An absent parser is not a content failure. It is the operator's, so it terminalizes the run
(`static_frontend_unavailable`) instead of burning the leaf's retry budget on something no source
can fix.

## A defect of my own, and where it was caught

The first version blanked interface spans one line too far, deleting the first statement after
`end interface` from the body — fail-open when that statement is the only assignment to the
`intent(out)` dummy. **The 365-file differential is structurally blind to it** (no in-tree model
declares an interface inside a body); the terminator × interface acceptance matrix caught it.

## Known defects fixed in stage B, not here

Found by four review rounds over the combined stack, all in this branch's code:

- recursive tree walk → `RecursionError` on deeply nested legal source, caught by the validator's
  own `RuntimeError` handler and reported as `schema_load_failed`, discarding every other violation
- the frontend-unavailable classification was a text scan over output that interpolates a
  leaf-chosen filename; forgeable three different ways, now an exit code
- statement labels: a labelled `DO` and a `FORMAT` among declarations were refused (legal, gfortran-clean)
- a grammar node-type rename would have produced zero procedures and zero errors — silent
- the CLI import fallback (the branch the conductor actually takes) and the canonical category
  order were unpinned
- packages not checked at launch, so their absence surfaced part-way into a billed run


seiya:
Superseded: collapsing the two stacked PRs into one.

This branch was stage A of a planned two-PR stack, but the review was run over the combined stack (`origin/main...feat/fortran-function-visibility`) rather than over this branch alone, and the fixes for defects found *in this branch's code* — a RecursionError reported as `schema_load_failed`, a classification channel a leaf could forge through its own filename, refusals of legal sources carrying statement labels, a silent failure on a grammar node-type rename — all landed on the stacked branch. So this tip is the pre-review state and is not independently mergeable.

Replaced by a single PR covering both stages, which is the unit that was actually reviewed and measured.

PR #67, opened 2026-08-14
Ask the registry which value the neutral core can author for, not which it knows

seiya:
Closes the **gates area** of the backend-boundary migration ledger (`TODO.md`, top item): the
deterministic gates that spelled the implemented set themselves now ask
`tools/backends/registry.py`. Until this landed, "add a backend without editing a gate" held for
the two signature gates only — `docs/BACKEND_BOUNDARY.md` §Operations Rules said so in as many
words.

## What changed

The registry gained the question these sites actually have. Four questions now, and asking the
wrong one is how each of this branch's fail-opens happened:

| Question | `None` / `True` when | Asked by |
| --- | --- | --- |
| `unsupported_reason` | the value is a declared member | naming, message building |
| `unimplemented_reason` | declared **and** something implements it | a gate deciding whether a run may carry the value |
| `provides(axis, value, capability)` | the **neutral core** does that named job for the value | a host-authorship dispatch |
| `unavailable_reason` | declared **and** extracted | code about to call into the backend package |

Wired: `_validate_toolchain_backend_supported` (per-axis capability, refusal clause taken from
the registry verbatim), the three `make`-quality-check sites via one shared predicate,
`_ir_is_m3c_physics`, the lint-evidence preset branch (`_LINT_ALLOWED_PRESETS` deleted — one
owner instead of two agreeing copies), the conductor's authorship predicates,
`orchestration_runtime.control_file_host_authored`, and `codegen_bundle`'s per-language tables
(moved to `tools/backends/language/fortran/bundle.py`).

## The plan was fail-open, and the correction is the point

The approved plan routed these sites through "declared **and** implemented". That is **worse than
the hard-coded pair it replaces**: the conductor's control-file writer emits GNU make syntax with
Fortran compile rules, so the predicate would answer True the day a second build system is
implemented AS A BACKEND, and route its nodes into the existing writer. Hence a *capability* —
declaring one asserts the inlined code exists now, and an extracted-but-capability-less backend
correctly falls to the documented leaf-authored path.

## Disclosure — defects this branch introduced, and review caught

Five commits: one implementation, four rounds of review fixes. The class descended, then
plateaued, which is why round 4 changed the instrument instead of adding another bespoke test.

1. **Two of the three mirrors of the authorship predicate were left behind.**
   `orchestration_runtime` (whose comment promises conductor and runtime "cannot disagree") and
   `_ir_is_m3c_physics` kept comparing against `(make, fortran)`. Measured: one capability
   declaration would double-own `src/Makefile` and silently disable `_validate_checks_source_files`
   for a node whose runner IS host-rendered.
2. **A witness that witnessed nothing.** The test added for the refusal clause probed `cmake`
   and asserted the implemented id `make` appeared — and `"cmake"` contains `"make"`. Both
   round-2 reviewers found it independently.
3. **Three checks of mine refused correct work**: registering a member before implementing it
   (the documented fail-closed default); the entire documented "Adding an axis" procedure; and a
   guard that flagged a legitimate mapping. My error direction on this branch was
   over-restriction, not under-restriction.
4. **An import-time refusal with the wrong blast radius.** A legitimate second language backend
   raised inside `import tools.codegen_bundle`, taking down the validator, conductor and
   pure_leaf — 374 tests unrun. Refusing was right; the radius was not.
5. **`str(None)` is `"none"`**, a live backend id on the `parallel` axis. The `or ""` guard
   existed and nothing observed it.
6. **Hardcoding `_resolved_makefile_host_authored = True` passed all 4718 tests** — only the True
   case was ever asserted. Extracting a predicate does not pin its call site.
7. **False prose**: `74 → 51` (measured 54); "a fourth site" for the third of three; a docstring
   claiming a function deliberately did not normalize while it did; a remedy clause naming the
   step this branch's own doc calls insufficient; a stale suite figure; and an assertion left
   searching for a string this branch had removed, so it could no longer fail.

## Verification

- Suite **4723 passed / 0 skipped** (4693 on `origin/main`).
- `ruff check tools/ mcp_servers/` — 83 errors, histogram **identical** to `origin/main`
  (measured on both sides).
- Sampled ratchet baseline 2,598 → **2,575**, regenerated in the commit that shed the tokens.
  Three counts ROSE and are naming, not knowing: `make-control-file` +2 in the validator, +2 in
  the conductor, `fortran` +1 in `BACKEND_BOUNDARY.md`.
- Review: 4 rounds, 9 subagents, 1 Codex pass (returned no actionable correctness regressions),
  and 6 sweeps of my own — hunk-level, mechanism-level, and before/after verification of every
  fix. Round 4 was a **witness census**: 233 decisions classified by execution, 111 mutations.

## Residue — stated, not closed

- **Production decisions left unwitnessed: 15**, each recorded with the input that would expose
  it. They need a second backend to exist before a test of them would mean anything: message
  wording and separator shapes, `unimplemented_reason`'s normalization (both callers lowercase
  first), the standalone-CLI import arm (`pragma: no cover`),
  `MAKE_QUALITY_CHECK_REQUIRED_LANGUAGES`'s `{c, cpp, mixed}` policy entries and
  `parallel_directives`'s declaration-only status (both filed against later ledger areas), and
  `max()` over a single language.
- **Structural limit**: 115 of the 233 decisions are test assertions, which cannot witness their
  own weakening — deleting one leaves a green tree by construction. No further round addresses
  that.
- **Two calls are provably redundant** (`_require_axis` inside `provides`; `implemented_backend_ids`
  in the `LANGUAGES` comprehension) and are kept as intent markers with comments saying so.
- **Not closed by this PR, deliberately**: the MCP `run_linter` preset chain spells the linter ids
  a fourth time, so registering a fifth linter is accepted by the validator and refused by the
  server. The `tools/` half is pinned by test; the MCP half is filed against the compiler/linter
  adapters area.

The convergence conditions were **not** met: no round returned zero functional findings. This is
a stop with a bounded, enumerated residue, not a claim of convergence.


seiya:

*(Misfiled. Review round 3 of the issue #181 pull request established that the `STALE_DEPENDENCY_IR_EXIT_CODE` channel landed in PR #88 (`1d1fa24`, merged `b79edf6`); the item's only `PR #N` mention — "the same direction recorded on PR #67 and PR #81" — is a comparison. The record for this item is https://github.com/seiya/atmofab/pull/88#issuecomment-5560285483. The verbatim body below is unchanged.)*

---

- **`STALE_DEPENDENCY_IR_MARKER` is forgeable from a leaf-chosen path, the same way the frontend marker was** (2026-08-14, medium — reproduced, pre-existing). `_gate_static_check` (`tools/workflow_conductor.py`) classifies a `post_generate` failure by searching the validator's whole output for `[stale-dependency-ir]`, and most violations interpolate a path the LEAF chose. A source file or directory whose name contains that literal therefore turns an ordinary content violation into `stale_dependency_ir`, which `GATE_FAILURE_TERMINAL` fail-closes — the run stops instead of warm-retrying `Generate.generate`, and the operator is pointed at a certified dependency IR that is fine.
  - Direction is fail-CLOSED, so this is misattribution and a burnt retry budget, not a gate bypass. Reproduced for the sibling marker `[fortran-structure-unavailable]` with a model source named `[fortran-structure-unavailable]_model.f90`, which produced the ordinary "present but must be named" violation and was classified `static_frontend_unavailable`.
  - The sibling was fixed by emitting the marker at the START of its violation and anchoring the conductor's match to the start of a violation line. That fix does NOT transfer as-is: `_validate_infrastructure_generated_signatures` interpolates its location first (`f"{loc}: {STALE_DEPENDENCY_IR_MARKER} …"`), so anchoring would have to move the marker there too, and that message's shape is pinned by its own tests.
  - Fix direction: emit the stale-dependency violation marker-first as well and anchor both scans, or replace text-scanning with a channel the leaf cannot write into (a distinct validator exit code, or a machine-readable sidecar the gate reads instead of stdout). The second closes the class rather than the instance, and there are only two instances.
  - Deliberately not done in the front-end branch: it predates it, `origin/main` scans the marker identically, and changing the message shape of an unrelated gate widens that branch's review surface.
  - **The same misattribution exists at `post_execute`, for BOTH markers, and there it has no terminal route to reach.** `validate --stage post_execute` runs the same `_validate_generate_outputs` → `_run_problem_model_gates` path (read, confirmed), so an absent Fortran front end or a stale dependency IR can be reported there too — but `_execute_inproc` classifies any non-zero gate rc as `post_execute_violation`, and every entry of `VALIDATE_EXECUTE_FAILURE_ROUTING` is a warm `("generate", "reuse")`. So a machine problem asks the leaf to re-author its source, which cannot fix it, and the retry budget burns. Not fixed with the Generate-side anchor because it needs a TERMINAL category in the execute routing table — a different subsystem's contract — and because the same one-line change would be needed for the stale-IR marker anyway. This is the third instance of the class, which is the argument for the structured-channel fix above rather than a third anchored scan. **`--stage pre_judge` is the fourth**: it dispatches through the same `validate()` path and its conductor call site scans for neither marker either (review, read).
  - **DONE (dedicated validator exit code, 2026-08-21)** — closed as a CLASS, by the structured channel this entry named as the second fix direction, not by a third anchored scan. `validate_pipeline_semantics.STALE_DEPENDENCY_IR_EXIT_CODE = 4` joins the front end's rc 3, and the decision is carried by the violation's TYPE (`StaleDependencyIRViolation(str)`, wrapping the single emit site) so `main()` maps it without reading its own output. The message is byte-identical — a `str` subclass compares equal to the string it wraps — so every existing message pin stayed green. All four conductor readers were enumerated and answered: `_gate_static_check` and `_execute_inproc` and `_post_judge_inproc` now classify on the code, and `classify_failure` grew `VALIDATE_EXECUTE_FAILURE_TERMINAL` (a separate constant from `GATE_FAILURE_TERMINAL`, pinned equal today) with the trial_meta read hoisted above the C2 counter so a machine failure never reopens `Compile`. The two remaining readers, `_build_inproc`'s post_build gate and `_compile_static_inproc`, carry a comment recording why they read no terminal code.
  - **Review-loop disclosure (4 rounds, 9 reviewers + 1 Codex pass, 18 fix commits on top of 7 implementation commits).** Class descent: R1 = missing witnesses and wrong numbers; R2 = a test whose docstring stated a rule it did not pin, plus a missed twin doc; R3 = **one mechanism with no witness at all** (see below) plus an over-refusal my own R2 fix introduced; R4 = zero functional defects from both reviewers, and a reviewer's own 11-mutant sweep fully killed. Codex ran once in R2 and returned clean, which is recorded as one independent pass consumed, **not** as convergence evidence.
    - **The defect my own fixes introduced twice, both times inside a test written to answer a review finding: over-refusal.** R2's rewrite of the exit-code distinctness row refused a legitimate ALIAS of an existing code; R3's fix of that had to state what it stops pinning. This is the same direction recorded on PR #67 and PR #81, and the countermeasure that worked again is the one those entries name — construct the legitimate input the check rejects, and default a floor to FILL rather than REFUSE.
    - **The worst finding was mine to have caught: exit code 3 had no behavioural witness.** Mutating `_main_dispatch`'s `return FORTRAN_STRUCTURE_UNAVAILABLE_EXIT_CODE` to `return 1` left 1555 rows green, while this change added three readers keyed on that code, a `--help` line and a RUNBOOK entry — and built a real-subprocess witness for the TWIN (rc 4) in the same commits. The test that carried the name said in its comment "this runs `main` in a real interpreter" and never called `main`. One test per OCCURRENCE of a rule, not per rule; and grep a test's body for the functions its docstring claims to drive.
    - **The twin-doc sweep took three rounds and two methods.** Reading and grepping found 2 of 6 sites; matching every `.md` under `os.walk` for `disposition`+`recoverable` found 5 and was written up as "the record of how many sites there are"; a reviewer using a different match expression found the 6th, stated with the word "severity". A census is a census only when it does not depend on one phrasing.
    - **Three rounds of numeric rot in one paragraph, twice inside a commit written to cure it.** Closed by changing the FORM, not the numbers: every measurement now names the commit it was taken at, and the paragraph states the property the count stands for. This is the fourth entry in this file to record the same lesson, which is itself the evidence that restating numbers does not work.
    - **Not met: the superior stopping condition** (the security axis producing nothing new for two consecutive rounds). R1's security reviewer found the unpinned dev category set, so the run is at most one clean round. Stopped on the ordinary condition — class descent plus both bounded-remainder proxies holding in R4 — with the remainder being descriptive accuracy, which ran 6-10 findings per round and did not fall.
  - **CORRECTION to the paragraph above.** "The same misattribution exists at `post_execute` … for BOTH markers" is **wrong for the stale-IR marker**, on this revision and on `origin/main`. The stale-IR violation has ONE emit site (`_validate_infrastructure_generated_signatures`) with ONE caller (`_validate_generate_outputs_for_generation`), which only the `post_generate` path reaches; `--stage post_execute` and `--stage pre_judge` reach `_validate_generate_outputs`, a different function that does not call it. The front-end marker's half of the claim IS correct — rc 3 reaches all three stages. The rc-4 branches at post_execute and pre_judge are therefore wired for a condition that cannot arrive today, and say so in their comments: the alternative is that the day the emit site gains a second caller, the failure arrives as a warm retry no leaf can converge on.
  - Verification: the channel is driven end to end by REAL subprocess tests (a helper-level assertion would stay green while an unwrapped `append` or a rebuilt violations list degraded production to rc 1), and the forgery is pinned at all three classifying readers: `_gate_static_check` replays the front-end marker's full four-payload defeat history at rc 1, while the post_execute and pre_judge rows replay two payloads each (one per marker), which is what those readers need — neither consults any text, so the rows pin the ABSENCE of a scan rather than sampling one. **Every measurement below names the commit it was taken at, because this paragraph rotted three times without that.** Each restatement was correct when written and was falsified by the next round's own commits — twice by a commit whose stated purpose was to restate it. Numbers about a moving branch are historical records, not claims about the present; written as records they cannot go stale, and a reader who needs today's figure runs the command.

Mutation-checked over the whole `origin/main...HEAD` range with `mutation_check.py`. **At `7c6d187`: 17 non-test hunks at git's default `-U3` (21 at `-U0`, 15 at `-U5` — a bare "N hunks" means nothing without the width), 10 killed, 6 prose-only, 1 the deletion half of the `trial_meta` read's MOTION** (reverting it restores a redundant second read of the same file below, which changes nothing — the moved copy above still runs). So every behavioural hunk is pinned: the `main()` mapping, each `elif`, and the placement of the terminal branch above the C2 counter each fail their own row when reverted. That property is what the count stands for, and it is the part that does not rot.

Suite **at `7c6d187`: 5064 collected on the branch against 5048 on `origin/main`; a node-id diff is 17 added, 1 removed** — the one removal is a test this loop RENAMED (`test_the_cli_answers_with_a_dedicated_exit_code`, whose comment claimed it drove `main` when it did not), so "0 removed" was refuted by a commit on this same branch. The `origin/main` control was taken in a `git worktree`, where 3 of those 5048 are location-dependent and NOT comparable (2 `ForbidBackendCredentialReadTests` rows resolve `../..` against the checkout, and one skips itself with "needs the checkout under $HOME"); all three pass or skip identically in the main checkout, so the collected counts are the honest comparison. (`7c6d187` is the last commit before this entry; the entry's own commit touches only `TODO.md`, which is outside both measurements.) `ruff check --statistics` per touched file is byte-identical to `origin/main` on all four: **1** in `workflow_conductor.py` (F841), **0** in `validate_pipeline_semantics.py`, **6** (4 F821 + 2 F401) and **1** (E741) in their tests — every one pre-existing (ruff 0.15.20, no in-tree ruff config). An earlier version of this line said 3/1/7/2, because the counter used to produce it was `ruff check <f> | grep -c "^[A-Z][0-9]*"` and `[0-9]*` matches zero digits, so `No fixes available` and its neighbours were counted as findings. The identical-to-`origin/main` half was never affected — the same wrong counter ran on both sides — but the absolute numbers were, which is the reason to read a count out of `--statistics` rather than out of a counter written for the occasion. The token baseline was regenerated twice, for **5** occurrences total: `tools/workflow_conductor.py` 75 -> 79 and `tools/validate_pipeline_semantics.py` 372 -> 373. Every one is the existing `FORTRAN_STRUCTURE_UNAVAILABLE_EXIT_CODE` identifier — an import and a comparison at each of the two new conductor read sites, and one interpolation in the `--help` epilog — i.e. a symbol the validator already exports, quoted in a neutral role rather than new technology knowledge. A reviewer verified the regeneration launders nothing: regenerating from a clean checkout reproduces the committed file byte-for-byte, and the 5 occurrences arrive on 5 added token-bearing lines with none removed (the baseline FILE's own numstat is 2/2 — two counts changed in place).
    - **The counter-argument, recorded because it is a judgement call and not a free pass**: `AGENTS.md` says flatly that adding to the measured neutral-core debt is forbidden, and 3 of the 5 were avoidable by spelling — a local alias bound once per site instead of importing the long name twice, and an epilog built from the value rather than the constant. Rejected deliberately: renaming an identifier to get under a substring counter satisfies the instrument and not the rule, and it is the same habit this repository names as the ratchet's own failure mode (teaching people to regenerate without reading the rule). The debt is real and is recorded here rather than hidden.
    - A **third** trip was NOT regenerated, because it was a genuine addition: a first draft of the `docs/RUNBOOK.md` recovery entry spelled the two parser distribution names, which `run_workflow.py`'s `REQUIRED_PYTHON_MODULES` already owns and prints. The entry points at that refusal instead. Three trips on one branch, all from ordinary prose, is the practical lesson: **the ratchet reads whole files, so any commit touching a scanned file has to run it** — one of the three was caught two commits late because the commit's verification ran only the test file for the module it changed.

PR #88, opened 2026-08-21
Route the stale-dependency-ir classification through a dedicated exit code (TODO:269-274)

seiya:
Closes the marker-forgery class recorded as TODO:269-274.

## The defect

`_gate_static_check` classified a `Generate.gate` failure as TERMINAL
(`stale_dependency_ir`, which `GATE_FAILURE_TERMINAL` fail-closes) by scanning the
validator's whole stdout for `[stale-dependency-ir]`. Violations interpolate a path the
LEAF chose (`f"{loc}: ..."`), so a model file named `[stale-dependency-ir]_model.f90`
turned an ordinary, warm-retryable naming slip into a terminal verdict — misattribution
in the fail-CLOSED direction, and a burnt billed run.

This is the twin of the front-end marker, which took **three narrowing text scans, each
defeated by one byte**, before an exit code closed it. TODO:269 named replacing the
channel as the fix that closes the class rather than the instance; that is what this does.

## The change

- `STALE_DEPENDENCY_IR_EXIT_CODE = 4`, alongside the rc-3 precedent.
- `StaleDependencyIRViolation(str)`: **the violation's TYPE is the channel**. A leaf writes
  bytes into a message, never a Python class. Subclassing `str` keeps the message
  byte-identical, so every existing message pin and every consumer that formats violations
  is unaffected. The single emit site wraps its `append`; `main()` maps
  `any(isinstance(...))` to rc 4, above `return 1` so a stale IR co-occurring with ordinary
  violations still routes terminal.
- **Six conductor readers, all answered.** `_gate_static_check`, `_execute_inproc` and
  `_post_judge_inproc` classify on the codes; `classify_failure` grows
  `VALIDATE_EXECUTE_FAILURE_TERMINAL` with the `trial_meta` read hoisted above the C2
  counter, so a machine failure never reopens Compile. `_build_inproc`'s post_build gate
  and `_compile_static_inproc` carry a comment recording why they read no terminal code —
  proved by call-graph closure, not by reading.
- Exit codes documented in `--help`, which `docs/CLI_REFERENCE.md` declares canonical for
  this CLI and which is the only channel a leaf has (it cannot read the module).
- Docs, a RUNBOOK recovery entry, and the TODO DONE record.

## Corrections to the ledger

TODO:274 recorded the same misattribution at `post_execute` "for BOTH markers". That is
**wrong for the stale-IR marker**: it has one emit site with one caller, reached only from
`post_generate`. The rc-4 branches at post_execute and pre_judge are therefore wired for a
condition that cannot arrive today, and say so — the alternative is that the day the emit
site gains a second caller, the failure arrives as a warm retry no leaf can converge on.
The front-end half of the claim is correct; rc 3 reaches all three stages.

## Verification

Measured at `7c6d187` and stated as such in the ledger, because this paragraph rotted
three times when it was written as a claim about the present:

- Suite **5064 collected / 5064 passed / 0 failed**, against 5048 on `origin/main`; node-id
  diff 17 added, 1 removed (a test this loop renamed).
- Mutation check over `origin/main...HEAD`: **17 non-test hunks at `-U3`, 10 killed, 6
  prose-only, 1 the deletion half of a code motion.** Every behavioural hunk is pinned.
- `ruff check --statistics` byte-identical to `origin/main` on all four touched files.
- The channel is driven end to end by **real subprocess** tests, because a type channel
  dies silently on any rebuild of the violations list.

## Disclosure: what the review loop cost

Four rounds, nine reviewers plus one Codex pass (clean, recorded as one independent pass
consumed, **not** as convergence evidence). 18 of the 26 commits are review fixes.

- **The worst finding was mine to have caught.** Exit code 3 had **no behavioural witness**
  — mutating its mapping to `return 1` left 1555 rows green — while this change added three
  readers keyed on it, a `--help` line and a RUNBOOK entry, and built a real-subprocess
  witness for the *twin* in the same commits. The test that carried the name claimed in its
  comment to run `main` and never called it.
- **Over-refusal entered twice, both times inside a test written to answer a review
  finding** — the direction this repository already records as my default error.
- **The twin-doc sweep took three rounds and two methods**: reading and grepping found 2 of
  6 sites; one `os.walk` match expression found 5 and was written up as a census; a
  reviewer using a different expression found the 6th.
- The token ratchet tripped three times, all from ordinary prose; two were neutral-role
  citations and regenerated, one was a real addition and was withdrawn.
- One commit message asserted something unrun and was amended message-only.

**The superior stopping condition — the security axis producing nothing new for two
consecutive rounds — was NOT met**; round 1's security reviewer found an unpinned dev
category set, so this run is at most one clean round. Stopped on the ordinary condition:
class descent, plus both bounded-remainder proxies holding in round 4 (a no-exclusions
reviewer returning zero functional defects, and a reviewer's own mutation sweep fully
killed). The remainder is descriptive accuracy, which ran 6-10 findings per round and did
not fall.

## Filed, not fixed

Two pre-existing leads recorded in TODO.md rather than pulled in: rc 4 has no launch-time
check where rc 3 does (so a stale IR costs one billed `Generate.generate` per occurrence),
and `classify_post_judge_violations` still decides a disposition by scanning bullet
basenames — filed with its reachability explicitly **not** established, so the next reader
runs the attack instead of inheriting a classification nobody earned.


seiya:
**Migrated from `TODO.md` (issue #181).** This item was recorded in `TODO.md` until commit `f88ad94` (main, 2026-09-06) and is reproduced below verbatim; the pull request for issue #181 removes it from that file. It was first filed on PR #67 by a destination rule that takes the first `PR #N` in the item body; review round 3 established that the `STALE_DEPENDENCY_IR_EXIT_CODE` channel landed in PR #88 (`1d1fa24`, merged `b79edf6`); the item's only `PR #N` mention — "the same direction recorded on PR #67 and PR #81" — is a comparison, so the record belongs here. Line-number citations (`L<n>`) and "above" / "below" references are as they stood in that file.

---

- **`STALE_DEPENDENCY_IR_MARKER` is forgeable from a leaf-chosen path, the same way the frontend marker was** (2026-08-14, medium — reproduced, pre-existing). `_gate_static_check` (`tools/workflow_conductor.py`) classifies a `post_generate` failure by searching the validator's whole output for `[stale-dependency-ir]`, and most violations interpolate a path the LEAF chose. A source file or directory whose name contains that literal therefore turns an ordinary content violation into `stale_dependency_ir`, which `GATE_FAILURE_TERMINAL` fail-closes — the run stops instead of warm-retrying `Generate.generate`, and the operator is pointed at a certified dependency IR that is fine.
  - Direction is fail-CLOSED, so this is misattribution and a burnt retry budget, not a gate bypass. Reproduced for the sibling marker `[fortran-structure-unavailable]` with a model source named `[fortran-structure-unavailable]_model.f90`, which produced the ordinary "present but must be named" violation and was classified `static_frontend_unavailable`.
  - The sibling was fixed by emitting the marker at the START of its violation and anchoring the conductor's match to the start of a violation line. That fix does NOT transfer as-is: `_validate_infrastructure_generated_signatures` interpolates its location first (`f"{loc}: {STALE_DEPENDENCY_IR_MARKER} …"`), so anchoring would have to move the marker there too, and that message's shape is pinned by its own tests.
  - Fix direction: emit the stale-dependency violation marker-first as well and anchor both scans, or replace text-scanning with a channel the leaf cannot write into (a distinct validator exit code, or a machine-readable sidecar the gate reads instead of stdout). The second closes the class rather than the instance, and there are only two instances.
  - Deliberately not done in the front-end branch: it predates it, `origin/main` scans the marker identically, and changing the message shape of an unrelated gate widens that branch's review surface.
  - **The same misattribution exists at `post_execute`, for BOTH markers, and there it has no terminal route to reach.** `validate --stage post_execute` runs the same `_validate_generate_outputs` → `_run_problem_model_gates` path (read, confirmed), so an absent Fortran front end or a stale dependency IR can be reported there too — but `_execute_inproc` classifies any non-zero gate rc as `post_execute_violation`, and every entry of `VALIDATE_EXECUTE_FAILURE_ROUTING` is a warm `("generate", "reuse")`. So a machine problem asks the leaf to re-author its source, which cannot fix it, and the retry budget burns. Not fixed with the Generate-side anchor because it needs a TERMINAL category in the execute routing table — a different subsystem's contract — and because the same one-line change would be needed for the stale-IR marker anyway. This is the third instance of the class, which is the argument for the structured-channel fix above rather than a third anchored scan. **`--stage pre_judge` is the fourth**: it dispatches through the same `validate()` path and its conductor call site scans for neither marker either (review, read).
  - **DONE (dedicated validator exit code, 2026-08-21)** — closed as a CLASS, by the structured channel this entry named as the second fix direction, not by a third anchored scan. `validate_pipeline_semantics.STALE_DEPENDENCY_IR_EXIT_CODE = 4` joins the front end's rc 3, and the decision is carried by the violation's TYPE (`StaleDependencyIRViolation(str)`, wrapping the single emit site) so `main()` maps it without reading its own output. The message is byte-identical — a `str` subclass compares equal to the string it wraps — so every existing message pin stayed green. All four conductor readers were enumerated and answered: `_gate_static_check` and `_execute_inproc` and `_post_judge_inproc` now classify on the code, and `classify_failure` grew `VALIDATE_EXECUTE_FAILURE_TERMINAL` (a separate constant from `GATE_FAILURE_TERMINAL`, pinned equal today) with the trial_meta read hoisted above the C2 counter so a machine failure never reopens `Compile`. The two remaining readers, `_build_inproc`'s post_build gate and `_compile_static_inproc`, carry a comment recording why they read no terminal code.
  - **Review-loop disclosure (4 rounds, 9 reviewers + 1 Codex pass, 18 fix commits on top of 7 implementation commits).** Class descent: R1 = missing witnesses and wrong numbers; R2 = a test whose docstring stated a rule it did not pin, plus a missed twin doc; R3 = **one mechanism with no witness at all** (see below) plus an over-refusal my own R2 fix introduced; R4 = zero functional defects from both reviewers, and a reviewer's own 11-mutant sweep fully killed. Codex ran once in R2 and returned clean, which is recorded as one independent pass consumed, **not** as convergence evidence.
    - **The defect my own fixes introduced twice, both times inside a test written to answer a review finding: over-refusal.** R2's rewrite of the exit-code distinctness row refused a legitimate ALIAS of an existing code; R3's fix of that had to state what it stops pinning. This is the same direction recorded on PR #67 and PR #81, and the countermeasure that worked again is the one those entries name — construct the legitimate input the check rejects, and default a floor to FILL rather than REFUSE.
    - **The worst finding was mine to have caught: exit code 3 had no behavioural witness.** Mutating `_main_dispatch`'s `return FORTRAN_STRUCTURE_UNAVAILABLE_EXIT_CODE` to `return 1` left 1555 rows green, while this change added three readers keyed on that code, a `--help` line and a RUNBOOK entry — and built a real-subprocess witness for the TWIN (rc 4) in the same commits. The test that carried the name said in its comment "this runs `main` in a real interpreter" and never called `main`. One test per OCCURRENCE of a rule, not per rule; and grep a test's body for the functions its docstring claims to drive.
    - **The twin-doc sweep took three rounds and two methods.** Reading and grepping found 2 of 6 sites; matching every `.md` under `os.walk` for `disposition`+`recoverable` found 5 and was written up as "the record of how many sites there are"; a reviewer using a different match expression found the 6th, stated with the word "severity". A census is a census only when it does not depend on one phrasing.
    - **Three rounds of numeric rot in one paragraph, twice inside a commit written to cure it.** Closed by changing the FORM, not the numbers: every measurement now names the commit it was taken at, and the paragraph states the property the count stands for. This is the fourth entry in this file to record the same lesson, which is itself the evidence that restating numbers does not work.
    - **Not met: the superior stopping condition** (the security axis producing nothing new for two consecutive rounds). R1's security reviewer found the unpinned dev category set, so the run is at most one clean round. Stopped on the ordinary condition — class descent plus both bounded-remainder proxies holding in R4 — with the remainder being descriptive accuracy, which ran 6-10 findings per round and did not fall.
  - **CORRECTION to the paragraph above.** "The same misattribution exists at `post_execute` … for BOTH markers" is **wrong for the stale-IR marker**, on this revision and on `origin/main`. The stale-IR violation has ONE emit site (`_validate_infrastructure_generated_signatures`) with ONE caller (`_validate_generate_outputs_for_generation`), which only the `post_generate` path reaches; `--stage post_execute` and `--stage pre_judge` reach `_validate_generate_outputs`, a different function that does not call it. The front-end marker's half of the claim IS correct — rc 3 reaches all three stages. The rc-4 branches at post_execute and pre_judge are therefore wired for a condition that cannot arrive today, and say so in their comments: the alternative is that the day the emit site gains a second caller, the failure arrives as a warm retry no leaf can converge on.
  - Verification: the channel is driven end to end by REAL subprocess tests (a helper-level assertion would stay green while an unwrapped `append` or a rebuilt violations list degraded production to rc 1), and the forgery is pinned at all three classifying readers: `_gate_static_check` replays the front-end marker's full four-payload defeat history at rc 1, while the post_execute and pre_judge rows replay two payloads each (one per marker), which is what those readers need — neither consults any text, so the rows pin the ABSENCE of a scan rather than sampling one. **Every measurement below names the commit it was taken at, because this paragraph rotted three times without that.** Each restatement was correct when written and was falsified by the next round's own commits — twice by a commit whose stated purpose was to restate it. Numbers about a moving branch are historical records, not claims about the present; written as records they cannot go stale, and a reader who needs today's figure runs the command.

Mutation-checked over the whole `origin/main...HEAD` range with `mutation_check.py`. **At `7c6d187`: 17 non-test hunks at git's default `-U3` (21 at `-U0`, 15 at `-U5` — a bare "N hunks" means nothing without the width), 10 killed, 6 prose-only, 1 the deletion half of the `trial_meta` read's MOTION** (reverting it restores a redundant second read of the same file below, which changes nothing — the moved copy above still runs). So every behavioural hunk is pinned: the `main()` mapping, each `elif`, and the placement of the terminal branch above the C2 counter each fail their own row when reverted. That property is what the count stands for, and it is the part that does not rot.

Suite **at `7c6d187`: 5064 collected on the branch against 5048 on `origin/main`; a node-id diff is 17 added, 1 removed** — the one removal is a test this loop RENAMED (`test_the_cli_answers_with_a_dedicated_exit_code`, whose comment claimed it drove `main` when it did not), so "0 removed" was refuted by a commit on this same branch. The `origin/main` control was taken in a `git worktree`, where 3 of those 5048 are location-dependent and NOT comparable (2 `ForbidBackendCredentialReadTests` rows resolve `../..` against the checkout, and one skips itself with "needs the checkout under $HOME"); all three pass or skip identically in the main checkout, so the collected counts are the honest comparison. (`7c6d187` is the last commit before this entry; the entry's own commit touches only `TODO.md`, which is outside both measurements.) `ruff check --statistics` per touched file is byte-identical to `origin/main` on all four: **1** in `workflow_conductor.py` (F841), **0** in `validate_pipeline_semantics.py`, **6** (4 F821 + 2 F401) and **1** (E741) in their tests — every one pre-existing (ruff 0.15.20, no in-tree ruff config). An earlier version of this line said 3/1/7/2, because the counter used to produce it was `ruff check <f> | grep -c "^[A-Z][0-9]*"` and `[0-9]*` matches zero digits, so `No fixes available` and its neighbours were counted as findings. The identical-to-`origin/main` half was never affected — the same wrong counter ran on both sides — but the absolute numbers were, which is the reason to read a count out of `--statistics` rather than out of a counter written for the occasion. The token baseline was regenerated twice, for **5** occurrences total: `tools/workflow_conductor.py` 75 -> 79 and `tools/validate_pipeline_semantics.py` 372 -> 373. Every one is the existing `FORTRAN_STRUCTURE_UNAVAILABLE_EXIT_CODE` identifier — an import and a comparison at each of the two new conductor read sites, and one interpolation in the `--help` epilog — i.e. a symbol the validator already exports, quoted in a neutral role rather than new technology knowledge. A reviewer verified the regeneration launders nothing: regenerating from a clean checkout reproduces the committed file byte-for-byte, and the 5 occurrences arrive on 5 added token-bearing lines with none removed (the baseline FILE's own numstat is 2/2 — two counts changed in place).
    - **The counter-argument, recorded because it is a judgement call and not a free pass**: `AGENTS.md` says flatly that adding to the measured neutral-core debt is forbidden, and 3 of the 5 were avoidable by spelling — a local alias bound once per site instead of importing the long name twice, and an epilog built from the value rather than the constant. Rejected deliberately: renaming an identifier to get under a substring counter satisfies the instrument and not the rule, and it is the same habit this repository names as the ratchet's own failure mode (teaching people to regenerate without reading the rule). The debt is real and is recorded here rather than hidden.
    - A **third** trip was NOT regenerated, because it was a genuine addition: a first draft of the `docs/RUNBOOK.md` recovery entry spelled the two parser distribution names, which `run_workflow.py`'s `REQUIRED_PYTHON_MODULES` already owns and prints. The entry points at that refusal instead. Three trips on one branch, all from ordinary prose, is the practical lesson: **the ratchet reads whole files, so any commit touching a scanned file has to run it** — one of the three was caught two commits late because the commit's verification ran only the test file for the module it changed.

PR #128, opened 2026-09-01
Rename the project from met-dsl to met-forge

seiya:
Renames the project `met-dsl` -> `met-forge`. Issue #127; the approved plan is that issue's first comment.

A clean cut: nothing in the tree answers to the old name. That is measured — see "The clean cut is real".

**Read the disclosure section before merging.** The branch narrows a live security guard off the directory where the operator's data still is, and the migration step that restores coverage is the operator's to run. It is a defense-in-depth reduction rather than an open hole, and nothing said so until round 3.

## Commits

One replacement class per commit, tree green at each, ordered so a longer match lands before a shorter one (K1 before K5, K3 before K4). K7–K9 are review fixes.

| | class | mechanical replacement | files | occ. / lines | hand-edited |
|---|---|---|---|---|---|
| K0 | glossary + identity lines | — | 3 | — | — |
| K1 | `.claude/skills/metdsl-*` | `metdsl-(enforcement-change\|review-loop)` -> `metforge-\1` | 27 | 85 / 77 | **0** |
| K2 | environment namespace | `METDSL` -> `METFORGE` | 64 | 773 / 687 | 3 |
| K3 | operator-private root | `\.met-dsl` -> `.met-forge` | 26 | 175 / 166 | 3 |
| K4 | prose (+ `pure-22`->`pure-23`) | `met-dsl` -> `met-forge` | 22 | 53 / 46 | 5 |
| K5 | identifiers | `metdsl` -> `metforge` | 21 | 80 / 78 | 2 |
| K6 | operator migration item | — | 1 | — | — |
| K7 | round-1 fixes | — | 1 | — | — |
| K8 | round-2 fixes | — | 2 | — | — |
| K9 | round-3 fixes | — | 4 | — | — |

Counts re-derived with a `git ls-tree -r` walk over each parent, `re.findall` for occurrences and a per-line search for lines. `git diff origin/main...HEAD --stat` = 104 files.

### Separating the machine from the hand

Each commit's hand-edited set is derived, not asserted: apply that commit's own sed expression to the PARENT of every file it touched and compare with the child. Whatever the replacement fails to reproduce is a hand edit, and that list is what the commit message claims.

The plan specified the REVERSE direction (undo the replacement on the child, compare with the parent). That is unsound and was replaced: reversing `met-forge` -> `met-dsl` on K4 also rewrites the occurrences K0–K3 had already created, and reported 12 files instead of 5.

## The two enforcement commits

**K2 — the writer/forwarder/reader triple.** writer `tools/run_workflow.py` -> forwarder `tools/orchestration_runtime.py:8241` `LEAF_ENV_ALLOWED_PREFIXES` (decides what reaches `bwrap --setenv`) -> reader `mcp_servers/build_runtime_server.py:93` `_WORKFLOW_MODE_ENV_VARS` (the MCP server inside the leaf). Had the writer moved alone, the forwarder would have dropped both names at the sandbox boundary and every capability gate in the leaf would have gone inert with nothing red. All three move in K2.

**K3 — one constant carries the guard.** Revert `tools/hooks/common.py:1590` alone, with the rest of K3 in place: `tools/tests/test_hooks_common.py` goes from the 2 standing failures to **21** — 19 rows across every spelling class the guard defends. Reproduced independently by three reviewers.

**32 obfuscated spellings on 29 lines were rewritten by hand**, because they exist so the guard fails closed on a command that does NOT spell the root literally — so none contains the substring the mechanical pattern matches. Each replacement was executed, not reasoned about: `{d..f}` because `d,e,f` holds `e`, `{c..g..2}` because `c,e,g` does. Two do not expand and are the same two the old spellings did not either (`.met-forg{e}`, a bash single-element-brace no-op the row pins deliberately; `.met-forg{0..999999999}`, over the expander's bound, caught as `.met-forg*`).

`.m$'e't-dsl` was listed in the plan as reachable mechanically and was not — `$'e'` splits the literal. It surfaced as a RED row, not a silent gap.

## The clean cut is real

Through a real server process (`mcp_servers/mcp_call.py`), with the OLD name as the control:

| run | result |
|---|---|
| no workflow env, `run_syntax_check` | `ok: true, return_code: 0` |
| `METFORGE_WORKFLOW_MODE=1`, `run_linter` | refused, naming `METFORGE_WORKFLOW_MODE` |
| `METDSL_WORKFLOW_MODE=1`, `run_linter` | **not refused**; the linter runs |

The third is the one that had to be run: it shows the gate keys on the new name ALONE, not that both names work because something still answers to the old one. Re-run independently in rounds 2 and 3.

## Disclosure — what this branch costs

**1. The Bash read guard stopped covering `~/.met-dsl` at the merge, and that is where the data still is.** The branch's one red-then-green result, found by the disclosure round and by nothing else — every other instrument compares HEAD against itself. Measured at `e04c35e`, driving `evaluate_common_policy` with workflow mode on against the real `$HOME`:

```
cat ~/.met-dsl/operator_tokens/o.txt                       origin/main BLOCK -> HEAD ALLOW
cat ~/.met-dsl/start_claims/x.lock                         origin/main BLOCK -> HEAD ALLOW
cat ~/.met-dsl/homes/<oid>/claude/.credentials.json        origin/main BLOCK -> HEAD ALLOW
cat ~/.met-dsl/homes/<oid>/claude/projects/<slug>/x.jsonl  origin/main BLOCK -> HEAD ALLOW
```

By `AGENTS.md` §"Development premises": the credentials and the old orchestrations' tokens buy a leaf nothing toward reporting its task done — an old token does not validate for the running `orchestration_id`, which is what `dismiss_violation` looks up — and are out of scope. The last row is not: a past run's leaf transcript, whose reading `docs/workflow/WORKFLOW_CORE.md` forbids precisely because it shortens the route to done.

**It is a reduction from two layers to one, not an open hole.** bwrap is mandatory (`_probe_bwrap_sandbox`: a host failing the probe is unsupported and the run fails closed) and `render_bwrap_command` is `--clearenv` plus explicit binds with no `--ro-bind / /`, so `~/.met-dsl` is not in a leaf's mount namespace. **That was established by reading the bind construction, not by driving a real launch** — the one layer in this verdict that was not executed.

Not fixed with an alias for the old root: the rename is a clean cut by decision, and a second protected root is dead weight the moment the operator migrates. `TODO.md` step 1 now leads with the gap and says the step is urgent rather than tidy.

**2. Two migration steps were missing, both destroying the audit trail of every pre-rename run.** `orchestration_meta.json#claude_workflow_home` is stored absolute and two orchestrations record `/home/seiya/.met-dsl/homes/<oid>/claude`; and the leaf-transcript slug is derived from the checkout path, so renaming the checkout orphans `.../claude/projects/-home-seiya-met-dsl` inside the durable homes. Now steps 5 and 6.

**3. The rename broke the reproducibility of a measurement record, and the plan's own policy is what did it.** `docs/GLOSSARY.md` §13 says a record dated before 2026-09 names things under the current spelling — right for prose, wrong for a PROCEDURE pinned to a named old commit, where the variable name IS the command. `tools/tests/conftest.py` carries a recipe taken at `165c26f`, which predates the rename; followed literally with `METFORGE_*` every row returns 0. Verified by checking `165c26f` out and running it both ways. Three such sites are now annotated.

**4. The old spelling deliberately survives in four files, not two**, as an earlier version of this PR said: `docs/GLOSSARY.md` §13, the `TODO.md` migration item, and the two annotated recipe records above.

**5. Defects the review fixes themselves introduced.** The round-1 reword added a false sentence to K3 ("the residual scan at HEAD is zero" — it never was, and never could be, given the deliberate records). Round 3 caught it. Rather than restate it a third time it was deleted, and the check now lives here with its method.

**6. `tools/run_workflow.py:3477`'s spelling of `METFORGE_ORCHESTRATION_ID` is pinned by nothing.** Round-0 mutation, escalated to the whole suite with an un-mutated baseline in the same worktree: byte-identical results. Left unfixed and disclosed rather than pinned — the leaf's id is authored by the conductor, the MCP gate's signal is an OR that `METFORGE_WORKFLOW_MODE` (which IS pinned) satisfies, and the hooks fail closed without an id, so no route was established by which a misspelling there produces a wrong outcome. It is a gap in this branch's own safety net, not a defect in the tree.

## Post-checks (at `3f5d16f`)

```
old name outside the four deliberate files ....... 0 lines
tracked filenames containing the old name ........ 0
standalone `DSL` ................................. byte-identical to origin/main
                                                   (38 lines, same line numbers, filtered
                                                    on the project name as the claim states)
new-name census .................................. METFORGE 688 / met-forge 221 / metforge 154
```

The plan's residue regex `\.met-d|met-ds|\.m\?t-|\[m\]et-|\.m\*t-|et-dsl` flags its own successful outcome — the `[m]`, `?` and `*` classes contain no `dsl`, so it matches the correctly rewritten `.[m]et-forge` / `.m?t-forge` / `.m*t-forge`. Anchored on the old suffix it returns only the deliberate files.

## Suite and lint

| | result |
|---|---|
| `origin/main` @ `0d88bf6`, primary checkout | 5555 passed, **2 failed** |
| `HEAD` @ `3f5d16f`, primary checkout | 5555 passed, **2 failed** |

Same two rows: `ForbidBackendCredentialReadTests::{test_folds_chained_relative_cd, test_a_variable_cd_target_folds_from_where_bash_lands}`. They need the checkout more than one level under `$HOME` and this one sits at exactly one. In a worktree elsewhere a different pair of the same class fails and a third skips — which is why every suite figure in these messages now names its checkout. `TODO.md`'s entry on this named one pair as "the pair"; K4 adds the measured correction that the set is a function of the depth.

ruff over the 48 changed `.py`, HEAD vs an `origin/main` worktree: **identical** — 1150 findings, same rule-code distribution, all pre-existing.

## Review loop

Round 0 plus three rounds; the default budget, not the cap.

| round | axes | outcome |
|---|---|---|
| 0 | mutation + reverse-substitution + post-checks | 1 unpinned hunk (disclosure #6) |
| 1 | security-bypass / correctness+doc-truth | security **0**; 7 reported, 6 real — all false evidence, incl. an operator step that already fails |
| 2 | security-bypass / mechanical (sonnet) / **Codex** | security **0**; Codex 1 (P2, operations); mechanical 5 reported, 4 real, 1 false positive |
| 3 | **disclosure** ×2 | the guard gap, 2 missing migration steps, a broken measurement recipe, 8 commit-message defects |

**Not converged.** The security axis produced nothing for two consecutive rounds, which is the superior stopping condition — but round 3 raised a NEW and more severe class (a guard coverage change) rather than a descending one, so class descent does not hold. Stopping is the budget's decision, not a claim that nothing is left.

Codex was launched once, in round 2, and returned one P2 on exactly the axis this loop expects it to cover: operations. Its first launch was rejected because `/codex:review` does not accept focus text; relaunched bare.

The sonnet-delegated mechanical axis spawned background agents despite an explicit prohibition and returned no result; asked to report only what it ran itself, it produced an honest "not run" accounting from which 4 real findings came. Recorded as one data point for that experiment.


seiya:
Superseded by #129, and closed by GitHub rather than by a decision: this PR's head branch `chore/127-rename-met-forge` was renamed to `chore/127-rename-atmofab`, the rename did not retarget this PR, and with its head ref gone the PR cannot be reopened.

The work is not lost and the name changed for a reason. `met-forge` collided with existing GitHub projects (`MetForge`), so the branch was regenerated as a patch series onto `main` under the name **`atmofab`** — the intermediate name is in no commit message and nothing in the tree answers to it. #129 carries the regenerated series (K0–K14), the per-commit forward-identity evidence that the new tree is the substitution image of `3f5d16f` plus an enumerated six-file hand-edit set, and three further review rounds.

`3f5d16f` and this PR's ten commits remain resolvable on `chore/127-rename-met-forge-superseded`.

PR #131, opened 2026-09-02
Close issue #127: record the operator migration as done, measured

### LINKED ISSUE #127: Rename the project from met-dsl to atmofab
The repository name `met-dsl` reads as "a project that defines a DSL", but what it holds is a workflow platform that generates, verifies, and certifies weather and climate kernels from a `spec`. The name is renamed to `met-forge`.

Scope: a full rename — prose, titles, `LICENSE`, the two development skill directories under `.claude/skills/`, the environment-variable namespace `METDSL_*`, the operator-private root `~/.met-dsl`, marker strings, temporary-directory prefixes, Fortran fixture identifiers, the GitHub repository, and the checkout directory name. A clean cut: no alias and no fallback for the old name. The standalone term `DSL` (the predicate DSL of `io_contract.test_predicates` and the language-neutral DSL identifier convention) is not touched.

The approved plan is posted as a comment below.

seiya:
The repository half of the `met-dsl` -> `atmofab` rename landed at `85793cd`. What kept issue #127 open was the eight-step operator migration in `TODO.md`, which is not reachable from a commit. It is done; this records the measurement.

**Verified on the operator's machine 2026-09-02**

| step | evidence |
| --- | --- |
| 1, 7 | `~/.met-dsl` and `~/.met-forge` gone; `~/.atmofab` mode **700**, all three subdirectories 700 |
| 2 | `origin` = `git@github.com:seiya/atmofab`; `gh repo view` = `seiya/atmofab` |
| 3 | `~/.claude.json`: `projects["/home/seiya/atmofab"].hasTrustDialogAccepted = true`, `githubRepoPaths["seiya/atmofab"]` set, no `met-dsl` key |
| 4 | checkout at `/home/seiya/atmofab`; `git worktree list` single main entry; old `~/.claude/projects/` and `~/.cache/claude-cli-nodejs/` slugs gone; no `METDSL_*` in any of the five shell rc files |
| 5 | both `orchestration_meta.json#claude_workflow_home` and both `owner.json#repo_root` re-pointed; `workspace/` deliberately not swept |
| 6 | both durable homes hold their transcripts under `claude/projects/-home-seiya-atmofab` |
| 8 | no-op here — nothing outside the repository parses the stdout envelope |

**The check that says steps 4-6 landed together.** `prune_workflow_homes.py::inspect_entry` reports both durable homes owner-verified (`verdict=deletable`) rather than `refused:unverifiable_owner` — the state a skipped step 5 would have produced, and the one that would have pointed the operator at `--allow-unverifiable`. The `rg1` suite-residue home was deleted rather than moved.

**Acceptance MET, without a billed run.** The criterion named the preflight roster probe's `atmofab-roster-home-` temp prefix as the one thing a real run reaches that the suite does not. That probe is unbilled and modelless, so the `preflight` subcommand alone answers it: `status=pass`, both `can_launch_*` true, `sandbox_enforced=true` / `bwrap`, `claude_leaf_tool_roster_classified` pass on CLI 2.1.258. The one non-`pass` row, `claude_features_list_available`, is advisory and outside the launch gate.

**Tree and suite.** `git grep -niE "met[-_ ]?dsl"` = 25 lines, every one a deliberate historical reference the `docs/GLOSSARY.md` §13 exception requires. Suite at `85793cd`: 5555 passed, 2 failed — the depth-dependent `ForbidBackendCredentialReadTests` pair `TODO.md:642` already records; the new checkout is one level below `$HOME` exactly as the old one was, so the rename neither caused nor cured them.

**Not closed by this PR**: the two pre-existing defects the rename surfaced (the operator-private root spelled four times; the suite writing into the real root) stay open in `TODO.md` — put out of scope by the approved plan, each its own change under `.claude/skills/atmofab-enforcement-change`.

Closes #127.

https://claude.ai/code/session_01QNjkCtcG3Lw8J7zBJ38ppH


seiya:
Added a second commit: the two pre-existing defects this rename surfaced are now issues of their own, and `TODO.md`'s citations point there instead of at #127, which this PR closes.

- #132 — the operator-private root is spelled four times and `operator_secret_root()` is only one of them
- #133 — the test suite writes operator tokens into the real operator-private root, and the guard covers only half of it

Two statements in those items also went stale when the migration ran, and are corrected here rather than left to read as current: "it is the state on disk today" (step 1 closed that window on 2026-09-02 — the split is now a demonstration of why the coupling matters, not a live gap), and the residue item's set arithmetic over three roots, two of which steps 1 and 7 deleted (kept as a dated record, marked un-re-takeable; the defect itself is untouched).

seiya:
Correction: this PR was merged at `e0b142d` at 00:29:53Z, before the second commit was pushed, so the `TODO.md` half described in the comment above is **not** in it. It is now PR #134.

PR #135, opened 2026-09-02
Fix the audit's false negative when run as a script, and close the class (#130)

### LINKED ISSUE #130: audit_orchestration.py run as a script reports "no dangling active_child window" when there is one
`tools/audit_orchestration.py` run the way `docs/RUNBOOK.md` documents it — as a script — silently loses two of its sections, and one of them turns into a FALSE NEGATIVE on the exact post-mortem the tool exists for.

## What breaks

`tools/orchestration_diagnostics.py` has two FUNCTION-LEVEL imports spelled `tools.`-absolute with no import shim:

- `tools/orchestration_diagnostics.py:430` (`_leaf_transcript_path`) — `from tools.hooks.common import claude_leaf_projects_roots`
- `tools/orchestration_diagnostics.py:454` (`_claude_projects_dir`) — the same import

The module's own MODULE-level import at `:56-59` carries the shim (`try: from leaf_usage ... except ModuleNotFoundError: from tools.leaf_usage ...`), as do the three at `tools/audit_orchestration.py:32-52`, with the comment that states the hazard: *"script run: `sys.path[0]` is `tools/` ; package import: repo root on path"*. The two function-level ones were missed, so under `python3 tools/audit_orchestration.py` — where `sys.path[0]` is `tools/` and the repo root is NOT on the path — they raise `ModuleNotFoundError: No module named 'tools'`. `tools/run_workflow.py:34-48` prepends the repo root for exactly this reason; `audit_orchestration.py` does not.

Both call sites are inside `audit()`'s two broad `except Exception` handlers (`tools/audit_orchestration.py:827-829` and `:833-839`), which are there to keep diagnostics from breaking the audit. They do that, and they also erase the failure.

## Consequences

**1. The dangling-launch section reports "no dangling window" when there is one.** `detect_dangling_active_child` itself has no bad import, so it correctly returns the open window; `build_launch_incident` then calls `_leaf_transcript_path`, which raises; `audit()` catches it and sets `launch_incident = None`; and `_render_launch_incident`'s `elif not snapshots` branch prints "No dangling active_child window detected and no captured incident snapshots." An operator following `docs/RUNBOOK.md:715` (`python3 tools/audit_orchestration.py --orchestration-id <id>`), which §launch-incomplete-recovery names as THE way to read this section, is told the window is closed.

**2. `--token-cost-from-transcripts` reports `token-cost collection failed`** for every run, via `aggregate_parent_usage` -> `_claude_projects_dir` at `tools/orchestration_diagnostics.py:669`. This is the milder half: the message at least says something failed, though it names neither the cause nor which half.

## Reproduction (measured 2026-09-02, this checkout)

Synthetic orchestration with an open window and nothing else — `active_child_agent_run_id.txt` and `active_children/<arid>.txt` set, no `child_returns/<arid>.txt`, no terminal `agent_runs.jsonl` row:

```
$ python3 tools/audit_orchestration.py --orchestration-id <oid> --repo-root <scratch> --format markdown
## Dangling launch (active_child window)

No dangling active_child window detected and no captured incident snapshots.
```

The same `audit()` call with the repo root on `sys.path`:

```
## Dangling launch (active_child window)

An open active_child window was found with no child return / terminal agent_runs row — the child launch never completed.

| field | value |
|---|---|
| child agent_run_id | `11111111-2222-3333-4444-555555555555` |
```

Consequence 2 reproduces against the two real durable homes under `~/.atmofab/homes/` (found while doing the issue #127 operator migration, steps 5b/6: the section stayed `token-cost collection failed` after the transcripts were verified resolvable, which is what led here).

## Why nothing caught it

`tools/tests/test_orchestration_diagnostics.py:14` imports `from tools import orchestration_diagnostics as diag`, and every other consumer reaches the module as a package too. The suite therefore never executes the script path, and no test invokes `tools/audit_orchestration.py` as a subprocess. This is a class, not two lines: any future function-level `tools.`-absolute import in a module reachable from a script entrypoint fails the same way, and a broad `except Exception` will hide it the same way.

## Fix sketch (not implemented)

1. Give the two imports the same shim the rest of the file uses, or hoist them to module level where the existing shim already covers the module — check the cycle that pushed them into function bodies first (`tools/hooks/common.py:4763` names both by name).
2. Better: have `audit_orchestration.py` do what `run_workflow.py:34-48` does and put the repo root on `sys.path` at the top, which closes the class for this entrypoint rather than these two call sites.
3. A witness: run `tools/audit_orchestration.py` as a SUBPROCESS in the suite over the synthetic open-window fixture above and assert the dangling section reports the window. An in-process test cannot see this bug.
4. Separately worth deciding: `audit()`'s two `except Exception` handlers should record what they swallowed (the exception type and message) into the rendered section, rather than reporting a clean negative. The false negative is what makes this severity, and it survives any fix that only repairs the imports.

Pre-existing; not introduced by the rename. Found during the issue #127 operator migration.

seiya:
Closes #130

## The defect

Run the way `docs/RUNBOOK.md` §launch-incomplete-recovery names as the way to reach that section — `python3 tools/audit_orchestration.py --orchestration-id <id>` — `sys.path[0]` is `tools/` and the repo root is nowhere on the path. The two function-body `from tools.hooks.common import claude_leaf_projects_roots` in `tools/orchestration_diagnostics.py` therefore raised `ModuleNotFoundError` inside `audit()`'s best-effort `except Exception`, and the audit reported an **OPEN active_child window** as `No dangling active_child window detected`, at exit 0.

Reproduced on the operator's own data: `orch_20260827T165705Z_af3800b5` has a genuinely open window. At `origin/main` the script path prints the clean negative; at HEAD it reports child `28db71fa-…`, `compile/generate`, with the transcript from `~/.atmofab/homes/`.

A per-import shim does not fix it — `tools/hooks/common.py` imports `tools.hooks.operator_safety` unconditionally, so `hooks.common` is unimportable bare however it is spelled. Putting the root on `sys.path` is the only fix.

## What changed

- **Bootstrap + hoist.** `audit_orchestration.py` gets the same absolute-first bootstrap `tools/orchestration_runtime.py:93-97` already uses; the bare-first shims are *not* kept, since once the root is on the path they would let `leaf_usage` and `tools.leaf_usage` coexist as two module objects. `orchestration_diagnostics.py` hoists its hooks import to module level, so a consumer that cannot import `tools` fails at import, where no `except Exception` can turn it into a false negative.
- **The class, not just the cause.** `audit()`'s three best-effort handlers record what they swallow in `diagnostic_failures`; the renderer prints UNKNOWN instead of a clean negative when a section did not run; `main()` exits 2.
- **A second route to the same false negative** (round 1): an orchestration id naming no directory raised nothing at all, so every collector read missing files as empty and the audit printed a full report of zeroes ending in the clean negative. Reachable from the RUNBOOK's own command, which passes no `--repo-root` — from any other cwd, over a real open window. Now `orchestration_found`, a banner naming both causes, UNKNOWN, exit 2.

## Review loop: 5 rounds, stopped at the CAP — this did NOT converge

Round 0 + 5 rounds (the cap), plus one Codex pass in round 2 (clean; not treated as a convergence signal). **Neither stopping condition was met.** The superior condition (two consecutive clean security rounds) has never been reached in any recorded loop here and was not reached now; of the two "bounded remainder" proxies, the mutation sweep is fully killed but **a blank-slate reviewer did not return zero functional defects** — round 5 found four items. They are fixed, but the loop is being stopped by budget, not by evidence that nothing is left.

**Class descent**: design hole (round 1) → missing witness (round 2) → holes in the guards themselves (rounds 3–5). It did not descend below that, which is the honest reason this says "stopped at the cap" rather than "converged".

### Defects the fixes themselves introduced — the majority, as usual here

| round | found inside | what |
|---|---|---|
| 2 | round 1's fix | `CollectorsDoNotSwallowTests` raised `RuntimeError` only, so it pinned one exception *shape*; `except OSError` survived |
| 3 | round 2's fix | `with contextlib.suppress(ImportError)` fell between the AST scan and the type sweep — and `ImportError` is the class #130 itself raised |
| 4 | round 3's fix | the swallow relocated one frame down, into `detect_dangling_active_child`, beat both guards |
| 5 | round 4's fix | the `_read_jsonl` injection covered only its **first** call site, so two swallows one statement apart survived — and the docstring claiming it held "whatever the code in between looks like" was **false** |
| 5 | rounds 3–4's fixes | two **over-refusals**: the guards rejected a context-preserving `raise ... from exc` and a legitimate lazy import, i.e. they pinned the *result* rather than the *rule* |

The mechanism was broken in six spellings across four rounds and the instrument was rebuilt twice. What holds now, all measured at the tip: primitive-level injection (`Path.read_text` / `is_dir` / `is_file` / `glob`) with a test-private exception type and a `__cause__`-chain check; `_blanket_swallows`, one rule over the whole detection path that forbids a non-re-raising broad handler; and a typed sweep over the detector call for the narrowly-typed suppress neither of the others sees. All six swallow spellings fail; three correct-code controls (broad re-raise with context, bare log-and-re-raise, an additional narrow `except FileNotFoundError`) stay green.

### False records I wrote and corrected in-branch

- `bc212ef` claimed the hoist's module-level half "is killed". Measured at that commit, reverting the *placement* left 144 green — only outright deletion was killed. Corrected in `03d24ef` (`rebase -i` is unavailable in this environment).
- A `TODO.md` entry said nothing produces `match_method` "in the tree". The grep was scoped to `tools/ skills/ docs/` without saying so; persisted snapshots do carry it, and the entry's "delete the clause" option would have dropped real evidence. Corrected in `eb52b9c`.
- That correction then said "three snapshots"; there are **five**, all carrying the key. The check printed a `[:3]` slice and I recorded the slice. Both round-4 reviewers found it independently. Corrected in `689e618`.
- `f0a589f` cites `tools/prune_workflow_homes.py:81-98`; the block runs 81-100.

## Measurements

- Full suite at the tip, primary checkout: **5588 passed, 2882 subtests, 2 failed** — `ForbidBackendCredentialReadTests::{test_folds_chained_relative_cd, test_a_variable_cd_target_folds_from_where_bash_lands}`, the depth-dependent pair that fails identically on `origin/main`.
- Mutation sweep `origin/main...HEAD`: 13 hunks, **every hunk pinned**.
- ruff **13 findings at HEAD and 13 at `origin/main`, identical breakdown** (EXE001×3, I001×9, PLW1510×1) over the changed source files, plus the two test files at 16/16. Measured with **ruff 0.16.5**; this repo has **no ruff config**, so the number is a function of that build's defaults and will not reproduce on another version.
- Real-corpus regression: all **41** orchestrations under `workspace/orchestrations/` exit 0 with `diagnostic_failures == []` and `orchestration_found == True`, in both formats and with `--token-cost-from-transcripts`. No healthy run is turned into an exit 2.
- Issue #130's second consequence: `orch_20260827T105454Z_99f63b7e` now attributes **64,302,365** child tokens where it used to say `token-cost collection failed`; `orch_20260827T165705Z_af3800b5` returns a genuine "no child usage located".
- **Undisclosed cost, now disclosed**: the hoist makes `tools.hooks.common` unconditional, taking `import tools.orchestration_diagnostics` from ~11 ms at `origin/main` to ~26 ms. `a7ee5f8` quoted the 28 ms absolute without saying it was new.

## Residue — deliberately not fixed here

Filed in `TODO.md`, not as issues (nobody has decided to do them):

1. `match_method` is rendered but the *current* code no longer produces it; the live route prints `None` while persisted snapshots carry real values.
2. `_load_json_if_dict` degrades a corrupt `orchestration_meta.json` to `{}` reaching neither integrity channel — #130's class by a third route, left out because the branch had already taken one scope increase.
3. Two siblings still carry #130's shapes: `prune_workflow_homes.py:81-100`'s bare-first shim, and four function-body `tools.` imports in `orchestration_runtime.py` — recorded with what is *not* measured about each, since this branch's evidence does not reach those files.
4. The report-wide loaders have no swallow pin, though they produce every number in the report. Shipped code fails loud there, so this is unpinned-but-correct, not a hole.
5. **A declared scope, not a task**: the three surviving checks are pinned to named functions and named primitives. A swallow in a function neither name covers, or a reader bottoming out in a primitive none of them patches, is outside all three. A strong claim would need both lists derived from the call graph instead of hand-maintained — a different change with its own evidence requirement, written down rather than attempted because the instrument had already been rebuilt twice.

Out of scope under `AGENTS.md` §Development premises, stated with what a leaf would gain: **nothing**. A leaf never reads this tool's output (nothing in `skills/` invokes it; it holds only the `Bash(python3 tools/audit_orchestration.py *)` grant), and every change here is fail-*louder*. The whole branch is about a false record an **operator** acts on. One round-5 finding — an unreadable `active_children/` producing a clean negative — is out of scope on the same premises: only the operator can create that state, and the operator owns the machine.

https://claude.ai/code/session_01SxmqMTSmKeXvAMPyWFMKZJ


seiya:
## Correction: two of the five residues listed above are not what this PR said they were

Measured after merge, at `ff3d7a9`. **The residue list in the PR body overstates the remaining work** — the same failure mode the body itself warns about, in the other direction. Both entries had been filed saying explicitly what was *not* measured; measuring them retires one and downgrades the other. Neither is an instance of issue #130's class.

**Residue 2 — `_load_json_if_dict` degrades a corrupt `orchestration_meta.json` silently: DOWNGRADED.**
The reachability the entry left open was "a crash mid-write truncates the file" — the very scenario the dangling-launch audit exists to diagnose, which would have made it the most reachable of the five. It is not reachable that way. All **15** writers touching `meta_path` in `tools/orchestration_runtime.py` go through `_write_json` (`:3151-3152`) → `_atomic_write_text` (`:3109-3149`): `tempfile.mkstemp` in the same directory, then `os.replace`. Zero non-atomic writers. That helper's own docstring says it exists so a reader "can transiently observe an empty or truncated file mid-write" cannot happen. So a corrupt meta needs real storage corruption or a hand edit, which `AGENTS.md` §Development premises puts out of scope as a path only the operator can reach. What stays in scope is the shape, not this file: the swallow still reaches neither integrity channel, so a *non-atomic* writer of some other `_load_json_if_dict` input would bring the silence back.

**Residue 3 — the two siblings still carrying #130's import shapes: RETIRED, both halves.**
- *No dual module identity.* `tools/prune_workflow_homes.py` driven through `runpy` with `PYTHONPATH` dropped and `sys.path[0]` set to `<repo>/tools` — the shape of a direct script run — then `sys.modules` dumped: `orchestration_runtime` is present under the **bare name only**, 14 `tools.*` modules are loaded, and the set of names loaded under **both** spellings is **empty**. Structural rather than lucky: the bare `orchestration_runtime` bootstraps the repo root itself at module level, so everything it imports is already `tools.*`, and nothing re-imports it dotted. (Limit: the run was `--help`, so only module-level imports executed — which is what decides identity, since they all run before argparse.)
- *The four deferred `tools.` imports in `orchestration_runtime.py` cannot fail the way #130's did.* Its bootstrap is at **module** level (`:92-97`), so the repo root is never missing by the time those function bodies run. #130's mechanism was a missing root. A genuinely absent module would still raise, but that is a different and much weaker risk than the one the entry was filed for.

What is left there is a style inconsistency — the shim is the spelling `tools/audit_orchestration.py` moved away from — not a defect.

**Still open, unchanged:** residue 1 (`match_method` rendered but no longer produced; the live route prints `None` today), residue 4 (the report-wide loaders have no swallow pin, though the shipped code fails loud there), and residue 5 (the declared scope: the three guards are pinned to hand-maintained function and primitive lists).

`TODO.md` is canonical for all of this and is being updated with the same measurements; these entries are kept rather than deleted, per that file's convention and because "this shape exists here" is still true — a deleted entry gets re-filed by the next reader.

https://claude.ai/code/session_01SxmqMTSmKeXvAMPyWFMKZJ

PR #140, opened 2026-09-02
One resolver per operator-private subtree, and the suite stops writing into the operator's own store (#132, #133)

### LINKED ISSUE #132: The operator-private root is spelled four times and `operator_secret_root()` is only one of them
Split out of #127 (found 2026-09-01 while doing the `met-dsl` -> `atmofab` rename; pre-existing, put out of scope by the approved rename plan). Recorded in `TODO.md`.

## What

`tools/hooks/common.py:1590` `operator_secret_root()` is the spelling the **read guard** has. Three writers build the same path from `Path.home()` instead:

- `tools/orchestration_runtime.py:19496` (`init_orchestration`) — `operator_tokens/`
- `tools/orchestration_runtime.py:10037` (`dismiss_violation`) — `operator_tokens/`
- `tools/run_workflow.py:1580` (`_claim_lock_path`) — `start_claims/`

`tools/hooks/common.py:1593-1596` states the rule for the SIBLING constant and says why — the side that CREATES and the side that FORBIDS reading must resolve the location identically — and the token and claim paths do not follow it.

## Why it matters

Nothing couples the four. Moving the root is four coordinated edits, and missing any one splits a writer from the guard **silently**.

The rename demonstrated exactly that: `operator_secret_root()` moved to `~/.atmofab` at commit `f5c6865` (K3) while the data was still under `~/.met-dsl`, deliberately and for one migration step's duration. Driving `evaluate_common_policy` with workflow mode on at `00fc29b`, four reads under the old root that were BLOCK at `origin/main` were ALLOW at HEAD for a leaf whose own `<orchestration_id>` is not that home's. That window is now closed — #127's migration step 1 moved the data — so this issue is about the coupling, not about a live gap on disk.

## Severity

Stated rather than assumed. A leaf gains nothing from either half being wrong: it cannot edit the tree, and the operator owns the machine. Under `AGENTS.md` §Development premises this is a **maintenance hazard on the operator's side, not a `leaf shortcut`**.

## Fix

Import `operator_secret_root` at the three writers, and add the test that the four agree.

It touches the capability gate's token path, so it is its own change under `.claude/skills/atmofab-enforcement-change`.

## Related

Same root cause, seen from the other side, as #133 (the test suite writing into the real operator-private root).

### LINKED ISSUE #133: The test suite writes operator tokens into the REAL operator-private root, and the guard built to stop exactly that covers only half of it
Split out of #127 (found 2026-09-01 while doing the `met-dsl` -> `atmofab` rename; pre-existing, deliberately not fixed there). Recorded in `TODO.md`.

## What

`tools/tests/conftest.py::_forbid_isolated_homes_in_operator_secret_root` fails any test about to resolve the isolated-homes root to the real root, and `ATMOFAB_WORKFLOW_HOMES_ROOT` gives `homes/` a redirect to resolve to.

`operator_tokens/` has **neither**. `tools/orchestration_runtime.py:19496` (`init_orchestration`) and `:10037` (`dismiss_violation`) each build `Path.home() / ".atmofab" / "operator_tokens"` directly, with no environment override and nothing in `conftest.py` watching. So a `pytest` run over `tools/tests/` creates the root if it is absent and leaves one file per fixture orchestration id in it.

**The redirect `homes/` does have is not the whole story either.** The guard is a `pytest` fixture wrapping `runtime._workflow_homes_root`, so it sees nothing driven OUTSIDE the runner — which is exactly what a review round does when it drives a function in a scratch tree. Measured 2026-09-01: `~/.met-dsl/homes/` held THREE entries, the third being `rg1` (`owner.json#repo_root = /tmp/tmp4ymg3k6z`, with a full `claude/` home) — the fixture id at `tools/tests/test_orchestration_runtime.py:16331`, landed *after* the guard did. So "covers only half" is the shape of the gap, not its extent: both halves leak, by different routes.

## Measurement

MEASURED 2026-09-01 by set arithmetic over the three roots that existed at the time. `~/.met-forge/operator_tokens` and `~/.atmofab/operator_tokens` — neither of which any workflow ever wrote, only that branch's suite runs — held the same **249** filenames, and that set was a SUBSET of `~/.met-dsl/operator_tokens`, which held **256**. So 249 of the live root's 256 were suite residue. The remaining 7 were 6 genuine run tokens plus `o.txt`.

The real `orch_<YYYYMMDD>T<hhmmss>Z_<hex>` shape does **not** separate the two groups: 7 filenames had it, and one of those (`orch_20260812T010203Z_ab12cd34.txt`) was inside the 249 — a fixture that spells itself like a live id. That is why the shape is a hint and the **set difference** is the measurement.

Reproduced independently by the rename's own migration record: `~/.atmofab` did not exist at 22:56; after ONE `python3 -m pytest tools/tests/` it existed, mode 755, with 249 fixture files in `operator_tokens/`.

## What it is not

Not a `leaf shortcut` and not a route to a wrong certification. A fixture id is not a live orchestration, `dismiss_violation` compares the token it reads against the one the operator holds, and the writer is the operator's own `pytest`, not a leaf. Under `AGENTS.md` §Development premises the leaf gains nothing here.

## What it is

The suite mutating a directory it declares off-limits, growing without bound, and re-creating the operator-private root behind the operator's back — which is what made #127's migration step 1 have to say "stat the target immediately before acting" rather than "run `mv`". A re-created root also comes back at the umask default: `_require_secure_home_ancestor`'s ancestor list records `~/.atmofab — mode not forced` (`tools/orchestration_runtime.py:16562`), while `:19501` chmods only `operator_tokens/`.

## Fix

- Give `operator_tokens/` the environment redirect `homes/` already has.
- Widen `_forbid_isolated_homes_in_operator_secret_root` to the **whole root** rather than one subdirectory.

It touches the capability gate's token path, so it is its own change with its own review under `.claude/skills/atmofab-enforcement-change`.

## Related

Same defect class as #132 (the operator-private root being spelled four times), seen from the other side.

seiya:
## What was wrong

`~/.atmofab` — the operator-private root — was spelled in four places, and only one of them was the Bash read guard's.

| Spelling | Site | Role |
|---|---|---|
| `(_home_dir() / ".atmofab").resolve()` | `tools/hooks/common.py::operator_secret_root` | the guard's only resolution point |
| `Path.home() / ".atmofab" / "operator_tokens"` | `init_orchestration` | token WRITER |
| same | `dismiss_violation` | token READER |
| `Path.home() / ".atmofab" / "start_claims"` | `run_workflow._claim_lock_path` | claim-lock WRITER |

**#132**: moving the root was a four-site coordinated edit; miss one and the writer and the guard split silently. The #127 rename had already done exactly that.

**#133**: `homes/` had an `ATMOFAB_WORKFLOW_HOMES_ROOT` redirect and a session guard; `operator_tokens/` and `start_claims/` had neither. A suite run wrote into the operator's real store — measured at `e0bae3d`, `HOME=$T python3 -m pytest tools/tests/test_orchestration_runtime.py` left **249** files in `$T/.atmofab/operator_tokens/`.

Both self-classified under rule 1-e as maintenance hazards, not `leaf shortcut`s: inside bwrap `$HOME` is forwarded as a value and the directory is not bound. **Two leaf shortcuts came out of the fix itself**, and both are described under Disclosure.

## What changed

- **One resolver per subtree**, all in `tools/hooks/common.py` beside the protected-read-root list: `operator_secret_root` / `operator_tokens_root` / `workflow_homes_root`, plus `run_workflow._start_claims_root` whose default hangs off the first. The three writers IMPORT them. `test_the_dot_atmofab_constant_is_spelled_once` holds the literal to one spelling, in that file.
- **The token store is a protected read root in its own right**, with `ATMOFAB_OPERATOR_TOKENS_ROOT`. It is a longer path than `~/.atmofab`, so the policy branch treats the two as one class — otherwise every token read would report under the backend-credential policy. The default block message is byte-identical to `origin/main`; a relocated store's names the path and the variable.
- **Creation-side refusals**, shared by the token store and the homes root: absolute or nothing; no containment with the checkout in either direction; the parent must exist; an existing path must be a directory. The resolvers stay validation-free because they feed the guard.
- **The isolated-homes ROOT is exempt from the containment drop**, like the token store and `~/.atmofab`.
- **The suite redirects all three subtrees per test and guards all three resolvers**, keeping PREVENT-not-DETECT: the guard wraps the resolver, so a mutant reaching past the redirect creates nothing.
- **`docs/RUNBOOK.md` §"The operator-private root" and `docs/HOOKS.md`** swept, with a rule 3-a check coupling the RUNBOOK section to the three constants resolved from code.

## Measurements (at `74c3fe1`, primary checkout `/home/seiya/atmofab`)

- Full suite **5632 passed, 0 failed** (2896 subtests). `e0bae3d`: **5595 passed**.
- `HOME=$T python3 -m pytest tools/tests/test_orchestration_runtime.py`: **249** files in `$T/.atmofab/operator_tokens/` at `e0bae3d`, **no `$T/.atmofab` at all** here. Same for `python3 -m unittest` with the three names unset.
- Real `~/.atmofab` unchanged by a full suite in the operator's own `$HOME`: **256** tokens, **40** claims, mode **700**.
- `ruff check` over the twelve touched `.py` files: **646**, against **645** at `e0bae3d`. The difference is one `EXE001` per new test module — the shebang convention 31 of the suite's existing modules already carry.
- End-to-end through `python3 -m tools.hooks.cli` in a subprocess: a relocated store denied naming the path and the variable; a file beside it allowed; `~/.atmofab/start_claims/x.lock` denied under the unchanged default wording.
- Both instruments RED at `e0bae3d`: the coincidence test (token in the operator's real store, claims root in the operator's real `~/.atmofab/start_claims`, no token-store entry in the guard) and the census (four members: `operator_secret_root`, `init_orchestration`, `dismiss_violation`, `_claim_lock_path`).

## Disclosure

**The review loop stopped at its cap (round 0 plus five), NOT at convergence.** The condition not met is the one this repository uses for a change that adds checking machinery: *a blank-slate review returning zero functional defects for two consecutive rounds*. Round 5's security axis returned zero and its correctness axis returned one. The superior condition — the security axis producing nothing new for two consecutive rounds — was also not met: round 2 was clean, round 3 was not, round 5 was clean. Rounds past a cap do find real defects; the budget is a decision to pay that cost, not a claim that nothing is left.

**Class descent, which is why stopping here is defensible.** Rounds 1–3 each produced something with a leaf on the other side of it. Rounds 4 and 5 produced nothing with a gain sentence.

**Two `leaf shortcut`s were introduced by this branch and found in review, both by the relocator it adds.**

1. **A token store inside the checkout was accepted** (Codex, round 2). `ATMOFAB_OPERATOR_TOKENS_ROOT=<repo>/spec/tokens` and `init_orchestration` wrote the dismiss-violation token there — measured. Every agentic leaf's `allowed_read_roots` contains `docs/` and `spec/` unconditionally and the Read tool never consults the protected roots, so a leaf could read the token and approve the `unauthorized_write_violation` its own substep produced. The RUNBOOK already warned against the configuration and named the wrong cost, because I had enumerated the Bash layer and not the manifest one. Refused now, in both containment directions, resolving symlinks. The premise itself is coupled: deleting `"spec/"` from the base read roots turns a test red.
2. **The homes root was still taking the containment drop** (round 3). With `ATMOFAB_WORKFLOW_HOMES_ROOT` containing the checkout, a leaf's `cat <homes-root>/<other-oid>/claude/projects/*/*.jsonl` was ALLOWED — a sibling orchestration's leaf transcript, the past-run state the workflow forbids. The justification in the code ("dropping one costs attribution, not enforcement, because this root still covers it") was true only at the default location. Same configuration axis PR #86 closed for the entry and left open for the drop.

**Defects the fixes introduced, in order.** An operator's `llm.yaml.rikyu` swept into a docs commit by `git add -A` (removed by rebase; `.gitignore` widened). A blind spot that MOVED rather than closed when I first pinned the conftest layers. A shared refusal helper handing the token store the *homes* wording. A regression against a sentence written on the same branch: `_start_claims_root`'s new `expanduser()` raising out of `_exclusive_claim`, which three places said degrades to "proceed". An over-refusal understated by a wide margin: a root AROUND the checkout blocks every Bash command in workflow mode, `echo hi` included, not "every recursive in-repo read".

**Records corrected on the branch, including two that cannot be fixed in the messages that carry them.** `b901d4a` says "`ruff check` over the three touched files unchanged" — FALSE, 527 → 529. `e252c01` says "6 unittest errors under a fake `$HOME` reproduce at `e0bae3d`" — NOT REPRODUCED, it is 3 at every revision; the paired "3 pytest failures" claim is right. Smaller: `b9f57f8` calls 223 "the five touched files" when that commit touches four, and `238708b` explains a net ruff +1 with an account of +2. Everything else in fifteen commit messages was re-measured by a dedicated axis and reproduced, including every mutant count and the two pre-fix security measurements.

**Recorded as unpinned rather than closed**, each proved vacuous by construction and not by corpus: `.absolute()` in `operator_tokens_root`; `mkdir(exist_ok=True)` vs `parents=True` under an override; the not-a-directory refusal's homes half (the ancestors loop refuses the same input three lines later); the middle conjunct of the relocated-message condition (entailed by the third); and the per-test loop in `isolated_homes_per_test_suite`, whose leak coverage is redundant with the module-level redirect. Each says so where it lives.

**Declared out of scope, with what the leaf would gain named as nothing**: `ATMOFAB_START_CLAIM_ROOT` is deliberately not a protected read root — a 0-byte advisory flock gets a leaf no closer to reporting a substep done. A relocated store pointed inside a backend credential home, and `=/` or `=/tmp`, are accepted and then fail closed. The `$(…)` fallback is weaker for a relocated store than for the default one — stated in the RUNBOOK, with the sandbox named as what closes it rather than this guard.

**Not coupled, and said rather than implied**: `docs/HOOKS.md`, the module comments, and the `--operator-token` argparse help. The rule 3-a check covers the operator-facing statement only.

## Filed in `TODO.md`, not fixed here

The private-root resolvers are not TOTAL for a `~account` override: all three call `expanduser()`, which raises `RuntimeError` when no such account exists. The two hooks resolvers fail CLOSED (`tools/hooks/cli.py` turns a raising hook into a BLOCK), so a leaf gains nothing; the cost is an operator whose typo refuses every Bash call without naming the variable. The claims one is closed at its caller — `_exclusive_claim` catches it — where "closed" means the driver no longer crashes, not that the operator is told. Not fixed because the fix is a decision, not a `try`: making a resolver total means saying what an unresolvable override resolves TO, and every answer either guards a path nobody writes to or guards a different tree from the one the writer refuses.

Closes #132
Closes #133

https://claude.ai/code/session_01SxmqMTSmKeXvAMPyWFMKZJ


seiya:
**Recorded from `TODO.md` (issue #181, PR-2).** This item stood in `TODO.md` in the form below until commit `58629e8` (main, 2026-09-07), and is reproduced verbatim. It is still OPEN: the pull request for issue #181 compresses it to its title, what is open, one measurement, and a link back here, so the file holds open work and the evidence that it is open and nothing else. The measurements, the review history and the rejected alternatives below are what left the file. Line-number citations (`L<n>`) and "above" / "below" references are as they stood in that file.

---

- **Left open by PR #140, filed rather than fixed: the private-root resolvers are not TOTAL for a `~account` override.** All THREE call `Path(override).expanduser()`, which raises `RuntimeError: Could not determine home directory` when the override is spelled `~someone` and no such account exists: `tools/hooks/common.py::operator_tokens_root`, `workflow_homes_root`, and `tools/run_workflow.py::_start_claims_root`. **The third is CLOSED at its caller and left open at the resolver**: `_exclusive_claim` catches `(OSError, RuntimeError, ValueError)` around `_claim_lock_path` so the claim degrades to "proceed" as its contract says (PR #140 round 4 — the `expanduser()` was added on that branch and the escape was a regression against a sentence the same branch wrote). So the note below about the hook layer converting a raise into a BLOCK, and the candidate fix in `tools/hooks/cli.py`, apply to the FIRST TWO ONLY. Whoever picks this up must not read it as two sites. "Closed at its caller" means the driver no longer crashes; it does NOT mean the operator is told — the degradation is silent, like the `fcntl` and unwritable-home arms beside it, so a mistyped claim root costs the cold-start serialisation with no signal. That is `_exclusive_claim`'s standing contract rather than something PR #140 introduced, and it is written down here because "closed" would otherwise read as more than it is. "Closed at its caller" means the driver no longer crashes; it does NOT mean the operator is told — the degradation is silent, like the `fcntl` and unwritable-home arms beside it, so a mistyped claim root costs the cold-start serialisation with no signal. That is `_exclusive_claim`'s standing contract rather than something PR #140 introduced, and it is written down here because "closed" would otherwise read as more than it is. Measured 2026-09-02 at PR #140's HEAD, and the same is true of `workflow_homes_root` at `e0bae3d` — the branch adds a second copy of the shape, not the shape. What follows is FAIL-CLOSED: `tools/hooks/cli.py` turns a raising hook into `_decision_error` and BLOCKS, so **a leaf gains nothing** (rule 1-e) and this is an operator-facing wart, not a hole — every Bash call in that shell is refused with `hook entrypoint failure: Could not determine home directory` and no mention of which variable caused it. Not fixed because the fix is a DECISION, not a `try`: making the resolver total means saying what an unresolvable override should resolve TO, and every answer either guards a path nobody writes to or silently guards a different tree from the one the writer refuses. The creation-side refusal cannot help — it never runs, because the resolver raises first. Candidate shape if this is ever picked up: catch in `tools/hooks/cli.py`'s entrypoint and name the offending `ATMOFAB_*` variable in the refusal, leaving the resolvers alone. That shape covers the two hooks resolvers; the claims one already has its answer at its own caller, and a fix that changes `expanduser` itself would have to keep `_exclusive_claim`'s degradation working. Found by a round-1 reviewer on PR #140.

Issue #183, opened 2026-09-06
Retire the meta-tests that pin a documentation or process preference, and move the two review-instrument tests beside their scripts


seiya:
Measured at `c131639` (main, clean tree), 2026-09-06.

## Problem

Eleven files under `tools/tests/` test documents, dev skills, the CI file, or the suite itself rather than product behaviour: 8,221 lines, 290 of the suite's 5,887 tests. Applying `AGENTS.md` §Development premises (a hole is in scope when a leaf that takes it gets closer to reporting its task done; its severity is the wrong certification or false record it produces) to each file gives three groups:

| group | files | lines | what removal costs |
|---|---|---|---|
| product record | `test_pure_prompt_contract_drift.py` (412), `test_operator_private_root.py` (745) | 1,157 | a stale `PURE_PROMPT_CONTRACT_VERSION` stamped into `bundle_meta.json` is a false record; two resolvers of the lock path can let two runs collide. KEEP. |
| suite integrity | `test_skip_reasons_are_declared.py` (827), `test_suite_environment_isolation.py` (924), `test_suite_private_root_harness.py` (92) | 1,843 | a product test that silently skips or reads the operator's environment hides a product defect; the first file's docstring records the instance ("skipped silently for weeks"). KEEP. |
| documentation / process preference | `test_backend_boundary.py` ratchet half (about 1,300 of 2,665; handled by #182), `test_skill_citations.py` (911), `test_ci_workflow.py` (721), `test_development_doc_sync.py` (522), `test_readme_sync.py` (184), `test_cli_reference_sync.py` (218) | 5,221 | a broken path in a dev skill, a README table one entry behind the code, a CI decision reversed. None is a wrong certification; a leaf gains nothing. |

Two more files, `test_mutation_check.py` (906) and `test_measure_claude_tool.py` (281), test scripts under `.claude/skills/` that no product code imports and CI never invokes (`docs/DEVELOPMENT.md:94` says the mutation check is "a procedure an author runs over their own diff"). They are tests of a development instrument living in the product suite. The scripts and the two dev skills themselves stay: `docs/DEVELOPMENT.md` §Record placement names `.claude/skills/*/references/` as the home of what a review loop taught, and removing them was considered and rejected.

[...]

Issue #175, opened 2026-09-06
spec_kind=profile is a compile-time selection policy: resolve it host-side at Compile and stop certifying it as generated code


seiya:
Measured at `c131639` (main, clean tree) by reading the two in-tree `profile` specs, every `profile` code path under `tools/` and `mcp_servers/`, and the profile orchestrations kept under `workspace*/`. Every number below was re-derived with the command shown; nothing is reasoned from the design documents alone.

## Problem

`docs/SPEC.md` §Design Policy gives a `profile spec` one role: "defines the `component` selection rules and parameter constraints". Both in-tree profiles (`dynamics_advdiff_profile_1d_upwind_center2_euler1`, `dynamics_shallow_water_profile_2d_rusanov_p0_ssprk2`) write that role down in §2 (select this profile by default when family and boundary match), §4 (no fallback), and §5 (record the resolution in `case.resolved.yaml`). No code realizes any of the three. The component set a profile "selects" is already pinned directly by the problem's `deps.yaml`, so the same three `component_id` + `version_constraint` pairs are stated three times per problem: problem `deps.yaml#components`, profile `deps.yaml#components` (reached through problem `deps.yaml#profiles`), and profile `controlled_spec.md` §1.

The operator's design statements (2026-09-06, recorded in the comments below) fix the role: for a part (a `component` or a `problem`) the scheme is decided at compile time and separate binaries per scheme are acceptable; runtime selection among schemes is the job of the driver of the assembled weather model, which selects among promoted subroutines. That driver will itself be built spec-driven by this workflow, so runtime switching inside one binary is a future node's concern, not a part's. A profile is therefore a compile-time selection policy for the node that adopts it, and nothing about it is code to certify.

Despite carrying no executable role, a profile is today a full certified node. It goes through Compile, Generate, Build and Validate like a component, costs five billed LLM leaves at attempt 1, produces a Fortran model that composes the three components (a second implementation of the problem's integration step) whose physics output no test judges, and is then compiled and linked into the problem's binary without a single reference from the problem's model, checks or runner. Its only measurable effect on the closure is negative: it is one more consumer whose closure bindings go stale on every component re-certification, and the three closure failures on record all happened at the profile node.

## Evidence (measured at c131639)

**No code executes the selection or fallback rule, and `case.resolved.yaml` has no producer.**

```
grep -rn '"profile"' --include="*.py" tools/ mcp_servers/ | grep -v /tests/
[...]

Issue #174, opened 2026-09-06
Tune / Promote: 198 lines of contract, 107 lines of unreachable write-root code and 378 lines of tests for two flows nothing can launch


seiya:
Measured on `c131639` (main, clean tree). `README.md:26` states that `Tune` and `Promote` are optional flows and that neither is implemented; `docs/WORKFLOW.md:21` and `docs/workflow/WORKFLOW_CORE.md:22,70` say their contracts are "handled in a separate plan". No such plan exists in this repository, in `TODO.md`, or in any issue (`gh issue list --state all --search 'optional flow'` returns 0), and the phase documents the runtime still cites for them were deleted on 2026-05-11. What remains is a contract with no owner and an enforcement branch with no caller.

## Problem

The repository carries two kinds of residue for the same two flows, and they disagree about what the flows are.

1. A deferred contract: three `SKILL.md` files, `docs/TUNING_WORKFLOW.md`, rows in `docs/AGENT_SKILLS.md` and `docs/workflow/LAUNCH_PROMPT_REFERENCE.md`, `official_releases: []` on every `spec_catalog.yaml` entry, an empty `releases/registry/component_catalog.yaml`, and about 61 cross-reference lines in 19 documents and schema files.
2. Live-looking code: a `promote` branch in `_write_roots_for_launch` that mints write roots under `releases/<spec_kind>/...` and on `spec/registry/spec_catalog.yaml`, a `tune` / `promote` branch in `_allowed_output_paths_for_launch`, and 13 tests that exercise only those branches.

The code is unreachable by construction: every entry point that mints or checks a capability asks `STEP_REQUIRED_CHILD_AGENT`, which names four steps, before the write roots are computed. The contract points at a plan that does not exist and at a phase document that was deleted. Neither half can be kept as is, and keeping both means every reader of the runtime has to re-derive that the `releases/` write root cannot be issued.

[...]

Issue #173, opened 2026-09-06
Delete the target-stack values the `Compile.static` gate can never admit: the `cppcheck` / `ruff` / `mixed` lint presets and the ten non-`make` build systems in the MCP server


seiya:
Measured at `c131639` (`main`, clean tree), 2026-09-06.

## Problem

The deterministic toolchain gate admits exactly one physical backend, `(build_system=make, language=fortran)`: `_validate_toolchain_backend_supported` (`tools/validate_pipeline_semantics.py:12311`) asks the registry for `build_execute`, `control_file` and `runner_render`, and `tools/backends/registry.py:255-262` declares them on `make` and `fortran` only; an `infrastructure` node is exempt on language but `_validate_published_surface` (lines 6492-6498) requires an extracted language backend equal to `_SIGNATURE_HELPERS_BACKEND_ID = "fortran"`. Downstream, `_require_make_build_system` (`tools/workflow_conductor.py:8486-8495`) refuses anything but `make` in `Build` and `validate.execute`.

The linter registry and the MCP server still carry code for values that gate refuses: two lint backends plus a composite, and ten build systems with a marker-file detector. Nothing in a run can select them; their only callers are their own tests. They cost CI an apt install, the runbook two version rows, and the tree a review loop (issue #120 spent 11 commits on the two lint presets on 2026-09-01 and left three ledger entries, two of which are about presets no node can reach).

One correction to the prior analysis: `tools/spec_input_gates.py` (112 lines) has no toolchain check; it holds `spec_id_length_violation` and `infra_dep_count_violation` only, and its line 11 says why ("there is no language to ask the registry about"). The rejection is at `Compile.static`, not at spec-input. The effect is the same; the location is not.

## Evidence (measured at c131639)

Reachability
- Lint preset selection: `_gate_lint_check` (`tools/workflow_conductor.py:9088-9114`) takes the preset from `_LINT_PRESET_FOR_LANGUAGE` (`tools/validate_pipeline_semantics.py:501-510`: c / cpp / c++ / cuda_c -> `cppcheck`, mixed -> `mixed`, python -> `ruff`) keyed by the IR's `toolchain.language`. The only caller-chosen path is standalone `run_linter` (`mcp_servers/build_runtime_server.py:1184-1201`, `mcp_servers/tools/run_linter.json`).
[...]

PR #184, opened 2026-09-06
Record the simplification program and the premise decisions it produced

seiya:
Records, inside the repository, the 2026-09-06 simplification survey, its independent verification, and the operator's decisions. Tracking issue: #167. Documents only; no code changes.

## What this records

- **`AGENTS.md` §Development premises: "No vendor lock-in".** No dependence on one LLM provider, one leaf transport, or one target-stack technology; a provider unused on the operator's machine is still part of the product. `docs/GLOSSARY.md` §12 defines `vendor lock-in`. Applied to reject #172 (delete `codex_cli`) and #173 (delete the lint presets and build systems no in-tree node selects; CUDA, C++ and Python targets are planned).
- **Tune / Promote are required future flows.** `docs/WORKFLOW.md` and `docs/workflow/WORKFLOW_CORE.md` no longer cite a "separate plan" that does not exist; the reserved surface stays (#174 rejected).
- **The role of `spec_kind=profile`.** `docs/SPEC.md` §Design Policy and `docs/GLOSSARY.md`: a compile-time selection resolved by the host for the adopting node, not a certified code node; runtime scheme selection belongs to the driver of the assembled model, itself a future spec-driven node (#175).
- **`docs/design/simplification_program.md`** (new, Status: decided): premise statements, method, the adopted issues with their landing order (#168 → #169 → #171 for the pure-leaf migration; #170, #175 → #183 independent), the rejected items with reasons, and the corrections the verification made to the survey.
- **`TODO.md`**: one item pointing at #167 with the 13 open issues.

## Checks

- `python3 -m pytest tools/tests/ -q -p no:randomly`: 5872 passed, 3168 subtests, 220 s at this tree.
- The document suites (`test_development_doc_sync`, `test_readme_sync`, `test_skill_citations`, `test_backend_boundary`, `test_cli_reference_sync`) pass; the boundary ratchet is unchanged because the edits to scanned documents use no backend token and `docs/design/` is outside the scan.

## After merge

Comment on #172 and #167 that the premise has landed in `AGENTS.md`.

https://claude.ai/code/session_014qe8fHjux8wjZoWzg15Hfs


seiya:
CI run 34016238175 failed on `LeafSpawnTest::test_claude_timeout_harvests_partial_output_when_the_pipe_stays_open` (`AssertionError: '' != 'partial answer\n'`), the genuine intermittent recorded in `TODO.md` (the #161 entry). This PR changes seven documents and no code on the leaf spawn path; the full suite passes on the development host at this tree (5872 passed, 3168 subtests). With this run the runner population is 6 failures in 20 runs (30%), against 5 of 19 before it.

Unverified hypothesis for whoever diagnoses it: the test feeds the fake `Popen`'s stdout from a scripted-pipe thread that writes `partial answer\n` and then holds the pipe open, with `_leaf_timeout_seconds` patched to 0.05 s and `LEAF_TERMINATE_GRACE_SECONDS` to 0.01 s. On a loaded runner the timeout and the harvest can complete before the writer thread has been scheduled, which yields exactly an empty stdout. If that is the cause, the fix is on the test side: gate the fake `wait()`'s `TimeoutExpired` on an event the writer sets after the partial line, so the timeout cannot fire before the line exists, rather than widening the budget. The development host has never reproduced it, so the check has to run on the runner.

Issue #185, opened 2026-09-06
The leaf-timeout partial-harvest test starves its own reader thread: a 20 ms window on a loaded runner, 6 of 20 CI runs red


seiya:
The genuine intermittent recorded in `TODO.md` (the #161 entry): `LeafSpawnTest::test_claude_timeout_harvests_partial_output_when_the_pipe_stays_open` fails on the CI runner with `AssertionError: '' != 'partial answer\n'`, 6 of 20 runner runs as of run `34016238175` (2026-09-06, PR #184, a documents-only change), never on the development host. This issue is the diagnosis the entry says the item exists for.

## Diagnosis (from the code path; the runner reproduction is the PR's job)

The test hands the conductor a fake `Popen` whose `stdout` is a real pipe already holding `partial answer\n` (the scripted writer thread runs from `_ScriptedStream.__init__`, long before `spawn_leaf` is called, so the bytes are in the kernel buffer) and whose `wait()` raises `subprocess.TimeoutExpired` IMMEDIATELY, without honouring its `timeout` argument.

`spawn_leaf` (`tools/workflow_conductor.py`, from `stdout_thread.start()`): starts the two reader threads, calls `process.wait(timeout=0.05)`, which raises at once, then `_terminate_leaf_process_group` (whose `_gone_within` calls `process.wait` again, raises again, returns False without sleeping), `process.wait(timeout=0.01)` (raises), and `_finish_streams`, which joins the readers for `LEAF_TERMINATE_GRACE_SECONDS` = 0.01 s twice and then sets `stop_reading` and copies `stdout_chunks`.

So the window in which the freshly started stdout reader thread must be scheduled, execute `os.read`, and append the block is about 20 ms of joins plus call overhead. On a loaded shared runner a new thread is not reliably scheduled within that; the copy then finds `stdout_chunks` empty. The data was in the pipe the whole time; it is the READER that is starved, not the writer. The always-identical symptom (empty stdout, stderr assertions never reached) and the dev-host non-reproduction (an idle machine schedules the thread at once) both fit.

Neither `--junitxml` nor the runner's core count enters this path; the `TODO.md` entry's caution about the report writer can be closed with this.

## Fix (test side; no product change)

[...]

PR #186, opened 2026-09-06
Make the leaf-timeout partial-harvest test wait for its own reader, not for luck

seiya:
Fixes the genuine intermittent recorded in `TODO.md` (the #161 entry) and diagnosed in #185: `LeafSpawnTest::test_claude_timeout_harvests_partial_output_when_the_pipe_stays_open`, red in 6 of 20 CI runs with `'' != 'partial answer\n'`, most recently on the documents-only PR #184.

## Cause

The test's fake `Popen.wait()` raises `TimeoutExpired` immediately, and it patches both graces to 10 ms, so the conductor spends about 20 ms between starting the reader threads and copying `stdout_chunks`. The partial line is already in the pipe; what must fit in that window is the new reader thread being scheduled, reading, and appending. A loaded runner misses it; an idle host never does. The report writer (`--junitxml`) and the runner's core count do not enter this path.

## Fix (test side; `tools/workflow_conductor.py` unchanged)

The fake `wait()` blocks (bounded, 10 s) until a wrapper around `wc._iter_stream_blocks` reports that the stdout reader has consumed the first block, then raises. The event is set when the consumer asks for the block after the first one, i.e. after `_drain_capped_stream` has appended it. The wedge path is exercised exactly as before; no budget is widened.

## Evidence (development host, this branch)

| run | result |
|---|---|
| unfixed test, every stream's first read delayed 200 ms | FAIL, `'' != 'partial answer\n'` (the runner's assertion) |
| fixed test, same delay | PASS |
| fixed test, 30 plain runs | 30 passed |
| `tools/tests/test_workflow_conductor.py` | 775 passed, 1111 subtests |

The starvation harness is a 15-line script kept outside the tree (it monkeypatches `_iter_stream_blocks` with a 200 ms delay and runs the one test under `unittest`); the PR description is its record.

## Completion criterion (from #185)

Ten CI attempts on this branch with zero failures of this test. Under the prior rate (6 of 20) ten consecutive greens would be a 3% event. Attempts are re-runs of this PR's workflow run; the count is reported in a comment when reached.

https://claude.ai/code/session_014qe8fHjux8wjZoWzg15Hfs


seiya:
Completion criterion of #185 met: ten attempts of this PR's workflow run (34016863727, attempts 1 to 10, started 06:34:58Z to 07:21:19Z on 2026-09-06) all `success`, each `5872 passed, 3168 subtests passed`, no failure of `test_claude_timeout_harvests_partial_output_when_the_pipe_stays_open`. Under the prior runner rate (7 of 21, 33%) ten consecutive greens would be about a 2% event. During the same window the unfixed test failed again on PR #184's rerun (run 34016238175, attempt 2).

